PR #5, opened 2017-06-25
Update Makefile to release and build for multiple platforms

luxas:
@DirectXMan12 PTAL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,49 @@
-all: build
+REGISTRY?=directxman12
+IMAGE?=k8s-prometheus-adapter
+TEMP_DIR:=$(shell mktemp -d)
+ARCH?=amd64
+ALL_ARCH=amd64 arm arm64 ppc64le s390x
+ML_PLATFORMS=linux/amd64,linux/arm,linux/arm64,linux/ppc64le,linux/s390x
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,13 +1,41 @@
-all: build
+REGISTRY?=directxman
```
DirectXMan12: I'm `directxman12` on dockerhub, too

Makefile
```diff
@@ -1,13 +1,41 @@
...
+endif
+ifeq ($(ARCH),arm)
+        BASEIMAGE?=armhf/busybox
+endif
+ifeq ($(ARCH),arm64)
+        BASEIMAGE?=aarch64/busybox
+endif
+
+all: build
 build:
```
DirectXMan12: I'd prefer to have the `build` target just be for normal `go building`, and have a `cross-build` target or something like that for this.

luxas: Actually this builds things for just one arch where amd64 is the default. However, I'll split it to `build` and `docker-build`

DirectXMan12:
I think this looks fine.  I *might* change `all` to just be `build` after the fact (haven't decided yet), but otherwise LGTM :+1:.  Thanks!

PR #6, opened 2017-06-26
Add vendor folder to git

luxas:
This will make it a lot more easier to consume the repo...
PTAL @DirectXMan12 

Only thing I did was `glide install -v` (with `-v` this time as you said :smile:)


DirectXMan12:
is the approach in #10 ok for you instead?  I'm not a huge fan of check in dependencies, and #10 makes it so the makefile runs the commands automatically when needed.

luxas:
it's ok, glide was much faster than I had expected

PR #9, opened 2017-06-26
Add RBAC and API Aggregation rules to example manifest

luxas:
Basically setting things up so you can access it via API Aggregation in an RBAC-enabled cluster as well
@DirectXMan12 PTAL

### DIFF
--- a/deploy/example-deployment.yaml
+++ b/deploy/example-deployment.yaml
@@ -1,34 +1,100 @@
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: custom-metrics
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: custom-metrics-apiserver
+  namespace: custom-metrics
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: custom-metrics:system:auth-delegator
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:auth-delegator
+subjects:
+- kind: ServiceAccount
+  name: custom-metrics-apiserver
+  namespace: custom-metrics
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: RoleBinding
+metadata:
+  name: custom-metrics-auth-reader
+  namespace: kube-system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: extension-apiserver-authentication-reader
+subjects:
+- kind: ServiceAccount
+  name: custom-metrics-apiserver
+  namespace: custom-metrics
+---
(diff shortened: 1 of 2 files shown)


deploy/example-deployment.yaml
```diff
@@ -1,34 +1,100 @@
...
+  namespace: custom-metrics
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: custom-metrics-resource-reader
+rules:
+- apiGroups:
+  - ""
+  resources:
```
DirectXMan12: we should have a note that the default policy restricts the resources to this set.  From a theoretical perspective, this could be `*`.  From a security perspective, that's probably sub-optimal.  What we really want is to be able to list the names of things w/o getting the actual contents.  This allows us to work with any resource, while limiting the attack surface in case of compromise.  It wouldn't be perfect (names can still leak important details), but it's *much* better than list itself.  @deads2k @liggitt either of you have opinions on this matter?

@luxas: also, I don't think we need the `get` permission, IIRC, just `list`.

deploy/example-deployment.yaml
```diff
@@ -39,3 +105,52 @@ spec:
...
+    targetPort: 6443
+  selector:
+    app: custom-metrics-apiserver
+---
+apiVersion: apiregistration.k8s.io/v1beta1
+kind: APIService
+metadata:
+  name: v1alpha1.custom-metrics.metrics.k8s.io
+spec:
+  insecureSkipTLSVerify: true
```
DirectXMan12: :-(.  I don't like doing this, although I suppose it's easier for people trying to deploy...

luxas:
> it makes it easier to recreate the different parts without kubectl yelling at you about existing resources.

I always do `kubectl apply` :)

Would it be possible to merge this as-is and then improve as we go?
I'm thinking about making this a Helm chart longer-term...

deploy/example-deployment.yaml
luxas: That requires creating a CA cert and using that to sign the serving cert and key.
Totally possible, but will get easier with a helm chart...

DirectXMan12:
> Would it be possible to merge this as-is and then improve as we go?

Yeah, that's fine.  Just do the README update I requested, and remove the `get` permission, and I'm fine with that.

> I'm thinking about making this a Helm chart longer-term...

Cool.

luxas:
> Just do the README update I requested

Hmm, where?

DirectXMan12:
sorry, it was buried in a blob of text:

> we should have a note that the default policy restricts the resources to this set. From a theoretical perspective, this could be `*`

DirectXMan12:
Merged, I'll add the README tweaks in a follow-up

luxas:
Yup, great to keep moving forward. Sorry, was buried in other work, this is my "do-for-fun" kind of stuff ;)

PR #10, opened 2017-06-26
Makefile Tweaks

DirectXMan12:
This sets up rules for ensuring the vendor directory is present and up-to-date in the Makefile, and makes `build` the default rule of `all`.

cc @luxas


Makefile
```diff
@@ -47,3 +48,6 @@ push: ./manifest-tool $(addprefix push-,$(ALL_ARCH))
 ./manifest-tool:
 	curl -sSL https://github.com/estesp/manifest-tool/releases/download/v0.5.0/manifest-tool-linux-amd64 > manifest-tool
 	chmod +x manifest-tool
+
+vendor: glide.lock
+	glide install -v
```
luxas: This assumes glide on host. I can install it for sure, but have everything I need in docker containers as a rule of thumb.

Makefile
```diff
@@ -23,11 +24,11 @@ ifeq ($(ARCH),s390x)
 	BASEIMAGE?=s390x/busybox
 endif
 
-all: docker-build
-build:
-	CGO_ENABLED=0 GOARCH=$(ARCH) go build -a -tags netgo -o /build/adapter github.com/directxman12/k8s-prometheus-adapter/cmd
+all: build
+build: vendor
+	CGO_ENABLED=0 GOARCH=$(ARCH) go build -a -tags netgo -o $(OUT_DIR)/adapter github.com/directxman12/k8s-prometheus-adapter/cmd
```
luxas: nit: Include $(ARCH) here as well? `$(OUT_DIR)/$(ARCH)/adapter`?

DirectXMan12: probably a good idea :-).  Will do.

Makefile
DirectXMan12: heh, I do the opposite -- I don't have Docker on my host system (it's in a bunch of VMs), so I tend to get a bit peeved when makefiles try to run docker.  I'll try to figure something out for this.

luxas: Hahah, let's meet somewhere in the middle then ;)

Makefile
```diff
@@ -47,3 +50,9 @@ push: ./manifest-tool $(addprefix push-,$(ALL_ARCH))
 ./manifest-tool:
 	curl -sSL https://github.com/estesp/manifest-tool/releases/download/v0.5.0/manifest-tool-linux-amd64 > manifest-tool
 	chmod +x manifest-tool
+
+vendor: glide.lock
+	glide install -v
+
+test: vendor
+	CGO_ENABLED=0 GOARCH=$(ARCH) go test ./pkg/...
```
luxas: This is one of the few cases where we actually don't need GOARCH set, as it will always run on the host platform which is automatically "correct" ;)

DirectXMan12: ack, will fix.

DirectXMan12:
Merging, will create a follow-up issue about running Glide from a Docker container.

PR #20, opened 2017-08-11
[docs] Add walkthrough

DirectXMan12:
This adds a walkthrough to the docs covering setting up Prometheus and
the adapter, as well as autoscaling a sample application on custom
metrics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,55 +37,33 @@ adapter talks to Prometheus and the main Kubernetes cluster:
 Presentation
 ------------
 
-The adapter gathers the names of available metrics from Prometheus at the
-specified interval.  Only metrics of the following forms are considered:
+The adapter gathers the names of available metrics from Prometheus
+a regular interval (see [Configuration](#configuration) above), and then
+only exposes metrics that follow specific forms.
 
-- "container" metrics (cAdvisor container metrics): series with a name
-  starting with `container_`, as well as non-empty `namespace` and
-  `pod_name` labels.
+In general:
 
-- "namespaced" metrics (metrics describing namespaced Kubernetes objects):
-  series with non-empty namespace labels (which don't start with
-  `container_`).
+- Metrics must have the `namespace` label to be considered.
 
-*Note*: Currently, metrics on non-namespaced objects (besides namespaces
-themselves) are not supported.
+- For each label on a metric, if that label name corresponds to
+  a Kubernetes resource (like `pod` or `service`), the metric will be
+  associated with that resource.
 
-Metrics in Prometheus are converted in the custom-metrics-API metrics as
-follows:
+- Metrics ending in `_total` are assumed to be cumulative, and will be
+  exposed without the suffix as a rate metric.
 
-1. The metric name and type are decided:
-   - For container metrics, the `container_` prefix is removed
-   - If the metric has the `_total` suffix, it is marked as a counter
-     metric, and the suffix is removed
-   - If the metric has the `_seconds_total` suffix, it is marked as
-     a seconds counter metric, and the suffix is removed.
-   - If the metric has none of the above suffixes, is is marked as a gauge
-    metric, and the metric name is used as-is
-
-2. Relevant resources are associated with the metric:
-   - container metrics are associated with pods only
-   - for non-container metrics, each label on the series is considered. If
-     that label represents a resource (without the group) available on the
-     server, the metric is associated with that resource.  A metric may be
-     associated with multiple resources.
-
-When retrieving counter and seconds-counter metrics, the adapter requests
-the metrics as a rate over the configured amount of time.  For metrics
-with multiple associated resources, the adapter requests the metric
-aggregated over all non-requested metrics.
-
-The adapter does not consider resources consumed by the "POD" container,
-which exists as part of all Kubernetes pods running in Docker simply
-supports the existance of the pod's shared network namespace.
+Detailed information can be found under [docs/format.md](docs/format.md).
 
 Example
 -------
 
-@luxas has an excellent example deployment of Prometheus, this adapter,
-and a demo pod which serves a metric `http_requests_total`, which becomes
-the custom metrics API metric `pods/http_requests`.  It also autoscales on
-that metric using the `autoscaling/v2alpha1` HorizontalPodAutoscaler.
+A brief walkthrough exists in [docs/walkthrough.md](docs/walkthrough.md).
+
+Additionally, [@luxas](https://github.com/luxas) has an excellent example
+deployment of Prometheus, this adapter, and a demo pod which serves
+a metric `http_requests_total`, which becomes the custom metrics API
+metric `pods/http_requests`.  It also autoscales on that metric using the
+`autoscaling/v2alpha1` HorizontalPodAutoscaler.
 
 It can be found at https://github.com/luxas/kubeadm-workshop.  Pay special
 attention to:
--- a/docs/format.md
+++ b/docs/format.md
@@ -1,0 +1,44 @@
+Metrics Format and Presentation
+===============================
+
+The adapter gathers the names of available metrics from Prometheus at the
+specified interval.  Only metrics of the following forms are considered:
+
+- "container" metrics (cAdvisor container metrics): series with a name
+  starting with `container_`, as well as non-empty `namespace` and
+  `pod_name` labels.
+
+- "namespaced" metrics (metrics describing namespaced Kubernetes objects):
+  series with non-empty namespace labels (which don't start with
+  `container_`).
+
+*Note*: Currently, metrics on non-namespaced objects (besides namespaces
+themselves) are not supported.
+
+Metrics in Prometheus are converted in the custom-metrics-API metrics as
+follows:
+
+1. The metric name and type are decided:
+   - For container metrics, the `container_` prefix is removed
+   - If the metric has the `_total` suffix, it is marked as a counter
+     metric, and the suffix is removed
+   - If the metric has the `_seconds_total` suffix, it is marked as
+     a seconds counter metric, and the suffix is removed.
+   - If the metric has none of the above suffixes, is is marked as a gauge
+    metric, and the metric name is used as-is
+
+2. Relevant resources are associated with the metric:
+   - container metrics are associated with pods only
+   - for non-container metrics, each label on the series is considered. If
+     that label represents a resource (without the group) available on the
+     server, the metric is associated with that resource.  A metric may be
+     associated with multiple resources.
+
+When retrieving counter and seconds-counter metrics, the adapter requests
+the metrics as a rate over the configured amount of time.  For metrics
+with multiple associated resources, the adapter requests the metric
+aggregated over all non-requested metrics.
+
+The adapter does not consider resources consumed by the "POD" container,
+which exists as part of all Kubernetes pods running in Docker simply
+supports the existance of the pod's shared network namespace.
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -37,55 +37,33 @@ adapter talks to Prometheus and the main Kubernetes cluster:
...
+a regular interval (see [Configuration](#configuration) above), and then
+only exposes metrics that follow specific forms.
+
+In general:
+
+- Metrics must have the `namespace` label to be considered.
+
+- For each label on a metric, if that label name corresponds to
+  a Kubernetes resource (like `pod` or `service`), the metric will be
+  associated with that resource.
```
luxas: What about a metric with more than one such label (e.g. `pod=foo,service=bar`)?
It will be discoverable from both sides, right?

docs/format.md
```diff
@@ -0,0 +1,44 @@
...
+
+- "container" metrics (cAdvisor container metrics): series with a name
+  starting with `container_`, as well as non-empty `namespace` and
+  `pod_name` labels.
+
+- "namespaced" metrics (metrics describing namespaced Kubernetes objects):
+  series with non-empty namespace labels (which don't start with
+  `container_`).
+
+*Note*: Currently, metrics on non-namespaced objects (besides namespaces
```
luxas: Do we have an issue for this?

README.md
luxas: nvm, clarified below

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+cluster are configured for autoscaling on custom metrics.  As of
+Kubernetes 1.7, this requires enabling the aggregation layer on the API
+server and configuring the controller manager to use the metrics APIs via
+their REST clients.
+
+Detailed instructions can be found in the Kubernetes documentation under
+[Horizontal Pod
+Autoscaling](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-custom-metrics).
+
+Make sure that you've properly configured Heapster with the `--api-server`
```
luxas: This flag isn't necessary AFAIK; it can use inclusterconfig instead...

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+flag, otherwise enabling custom metrics autoscaling support with disable
+CPU autoscaling support.
+
+### Binaries and Images ###
+
+In order to follow this walkthrough, you'll need container images for
+Prometheus and the custom metrics adapter.
+
+Both can be found on Dockerhub under `prom/prometheus` and
```
luxas: or use the CoreOS prometheus-operator

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+it `prom-cfg.yaml`, for example), and then create a ConfigMap containing
+it:
+
+```shell
+$ kubectl create configmap prometheus --from-file=prometheus.yml=prom-cfg.yaml
+```
+
+You'll be using this later when you launch Prometheus.
+
+### Launching the Deployment ###
```
luxas: Can we break out configuring Prometheus and deploying it from this section and just say: If you want to go the easy way; run these three commands and use the prom operator. If you want to go the custom way, look at this doc.

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+      # make your configmap available as a volume, so that you can
+      # mount in the Prometheus config from earlier
+      - name: config-volume
+        configMap:
+          name: prometheus
+```
+
+</details>
+
+Save this file (for example, as `prom-adapter.deployment.yaml`).  Now,
```
luxas: I wouldn't bundle prometheus and the adapter in the same pod in this walkthrough.
Deploy those separately

Here, the "Deploy the Custom Metrics Adapter for Prometheus" topic should start

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+standard Kubernetes addon API server.  This means that you'll need proper
+API server certificates for it to function properly.  To learn more about
+which certificates are needed, and what they mean, see [Concepts: Auth and
+Certificates](https://github.com/kubernetes-incubator/apiserver-builder/blob/master/docs/concepts/auth.md).
+
+Once you've generated the necessary certificates, you'll need to place
+them in a secret.  This walkthrough assumes that you've set up your
+cluster to automatically inject client certificates and CA certificates
+(see the above concepts documentation).  Make sure you've granted the
+default service account for your namespace permission to fetch the
```
luxas: I wouldn't recommend this :)
Create a dedicated SA for the adapter and bind this Role to that SA.

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+authentication CA ConfigMap:
+
+```shell
+$ kubectl create rolebinding prom-ext-auth-reader --role="extension-apiserver-authentication-reader" --serviceaccount=prom:default
+```
+
+Then, store your serving certificates in a secret:
```
luxas: Care to provide some openssl command how to gen these certs?

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+
+```shell
+$ kubectl create secret tls serving-cm-adapter --cert=/path/to/cm-adapter/serving.crt --key=/path/to/cm-adapter/serving.key
+```
+
+Finally, you'll need to make sure that the default service account for
```
luxas: again, dedicated SA

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+    # use your serving certs
+    - --tls-cert-file=/var/run/serving-cert/tls.crt
+    - --tls-private-key-file=/var/run/serving-cert/tls.key
+    # Prometheus is running in the same pod, so you can say your Prometheus
+    # is at `localhost`
+    - --prometheus-url=http://localhost:9090
+    # relist available Prometheus metrics every 1m
+    - --metrics-relist-interval=1m
+    # calculate rates for cumulative metrics over 30s periods.  This should be *at least*
+    # as much as your collection interval.
```
luxas: nit: collection interval for Prometheus.

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+  group: custom-metrics.metrics.k8s.io
+  version: v1alpha1
+  # these control how the aggregator prioritizes your registration.
+  # it's not particularly relevant in this case.
+  groupPriorityMinimum: 1000
+  versionPriority: 10
+  # finally, this points the aggregator at the service for your
+  # API server that you created
+  service:
+    name: prometheus
```
luxas: I prefer to deploy this aggregator in the `custom-metrics` ns or something like that, and have the service named `api`

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+```
+
+### Double-Checking Your Work ###
+
+With that all set, your custom metrics API should show up in discovery.
+
+Try fetching the discovery information for it:
+
+```shell
+$ kubectl get --raw /apis/custom-metrics.metrics.k8s.io/v1alpha1
```
luxas: hah, I didn't actually know about this, very cool :)

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+
+This sample application provides some metrics on the number of HTTP
+requests it receives.  Consider the metric `http_requests_total`.  First,
+check that it appears in discovery using the command from [Double-Checking
+Yor Work](#double-checking-your-work).  The cumulative Prometheus metric
+`http_requests_total` should have become the custom-metrics-API rate
+metric `pods/http_requests`.  Check out its value:
+
+```shell
+$ kubectl get --raw "/apis/custom-metrics.metrics.k8s.io/v1alpha1/namespaces/prom/pods/*/http_requests?selector=app%3Dsample-app"
```
luxas: The ns should be default here, right?

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+```
+
+It should be zero, since you're not currently accessing it.  Now, create
+a few requests with curl:
+
+```shell
+$ curl http://$(kubectl get service -o jsonpath='{ .spec.clusterIP }')/metrics
```
luxas: can use [`hey`](https://github.com/rakyll/hey) here as well

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+metrics sourced from the adapter.
+
+Prerequisites
+-------------
+
+### Cluster Configuration ###
+
+Before getting started, ensure that the main components of your
+cluster are configured for autoscaling on custom metrics.  As of
+Kubernetes 1.7, this requires enabling the aggregation layer on the API
```
luxas: Tell about the args that are required on the API server and the certs that have to be provided there in order for this to work. You can link to the API Aggregation docs as well, and tell users that kubeadm does this OOTB :)

docs/walkthrough.md
```diff
@@ -0,0 +1,500 @@
...
+server and configuring the controller manager to use the metrics APIs via
```
luxas: Tell me which arg, and how you can enable it with kubeadm to get going easily:
```yaml
kind: MasterConfiguration
apiVersion: kubeadm.k8s.io/v1alpha1
controllerManagerExtraArgs:
  horizontal-pod-autoscaler-use-rest-clients: "true"
  horizontal-pod-autoscaler-sync-period: "10s"
apiServerExtraArgs:
  runtime-config: "api/all=true"
```
```
kubeadm init --config=kubeadm.yaml
```

docs/walkthrough.md
DirectXMan12: that flag just say "run Heapster using the standard API server code as opposed to the home-grown go-restful code"

docs/walkthrough.md
DirectXMan12: I haven't configured that before, and, at least for this walkthrough, we depend on specific relabeling (pod and namespace).

docs/walkthrough.md
DirectXMan12: I've linked to a document that provides more details on how to do it (the "Concepts" doc on auth and certs).  I don't want to have that info reproduced everywhere, because it makes it harder to update.

docs/walkthrough.md
DirectXMan12: I provide a link below ("Detailed instructions can be found in the Kubernetes documentation...").  I can provide a link to the `kubeadm` docs as well, if there is one.  I don't use `kubeadm`, so I wasn't familiar with the config required.

DirectXMan12:
A couple of comments applied throughout, so I'll respond here:

> Deploying the adapter alongside Prometheus

The reason I suggest this in the walkthrough is that Prometheus does not support auth of any kind (even cert auth).  Ergo, it's not uncommon to set the Prometheus listening address to localhost, and then deploy behind an auth proxy (like https://github.com/openshift/oauth-proxy).  Since Prometheus doesn't support any auth, the adapter doesn't yet support any auth when talking to Prometheus, in those configurations, you'd want to run the adapter as a sidecar.

> The Operator

I'm not familiar with configuring the Operator, and we need certain relabeling, so I used this.  I'd take a PR for using the Operator and splitting the manual config into it's own file to avoid this getting too long.

> The default SA

If Prometheus is in its own NS, it's not too big a deal, but sure, I can add that in.

DirectXMan12:
addressed the easy comments.  We can deal with adding kubeadm docs and operator instructions later

luxas:
Yeah, sorry, have very limited time now for reviews due to exams...

PR #26, opened 2017-11-29
Fix certificate paths in deploy manifests

simonpasquier:


DirectXMan12:
Hmm... our documentation explicitly says to use `serving.crt` and `serving.key`: https://github.com/DirectXMan12/k8s-prometheus-adapter/blob/c3ad5ef0ea3a12d99ded19747994c1f42f080c40/deploy/README.md.

Can you clarify the issue that you had that prompted this change?

simonpasquier:
Ah in fact I mixed information from `deploy/README.md ` and from `docs/walkthrough.md`. The second document says to run "`kubectl -n prom create secret tls serving-cm-adapter ...`" (which generates `tls.crt` and `tls.key` secret keys) while the second one assumes generic secrets. This is where my confusion comes...

I'm closing the PR. Maybe some clean up is required for the walk-through document?

DirectXMan12:
yeah, I'm always willing to accept PRs to clarify things that aren't clear :-)

PR #33, opened 2018-01-23
Use rm -rf instead of sudo when removing the temp dir

vainu-arto:
The directory is owned by the user running make, but some files in it are owned by root since they are created by programs running within a docker container running as root. This doesn't matter, unlinking files is a write operation on the directory that contains them. Thus write access to the root-owned files isn't required.


DirectXMan12:
good catch, yeah, we don't need the sudo there.

PR #47, opened 2018-02-21
Fix serving cert path config in deployment spec

aschepis:
The volume was mounted at `/var/run/serving/certs` but the container was sending an arg to look at `/var/run/serving-cert` (missing the trailing `s`)


DirectXMan12:
thanks!

PR #49, opened 2018-02-22
Add missing `create` on api registration

aschepis:
when creating the custom APIService the example was missing the `create` portion of the `kubectl` command.


DirectXMan12:
good catch.  Thanks!

PR #53, opened 2018-03-11
Add support for auth when connecting to Prometheus

### LINKED ISSUE #52: the adapter doesn't support access to prometheus which enabled ssl/tls
I checked the codes, that adapter use http.DefaultClient to access prometheus instance and it doesn't support https client to access ssl/tls enabled prometheus. It is a gap here if prometheus behind a proxy or 
prometheus enable ssl/tls according to official guideline [Why don't the Prometheus server components support TLS or authentication? Can I add those?](https://prometheus.io/docs/introduction/faq/#why-don't-the-prometheus-server-components-support-tls-or-authentication?-can-i-add-those?) 

This is a common case that prometheus in production, do we have any plan for this? Thanks in advance.

huxiaoliang:
fix #52

Currently, the adapter use http.DefaultClient to access prometheus
instance and it doesn't support https client to access ssl/tls is
enabled prometheus. It is a gap here if prometheus behind a proxy
or prometheus is enable ssl/tls according to official guideline.
This is a common case that prometheus in production, this patch
will fix the gap described above.

### DIFF
--- a/cmd/adapter/app/start.go
+++ b/cmd/adapter/app/start.go
@@ -81,12 +81,50 @@ func NewCommandStartPrometheusAdapterServer(out, errOut io.Writer, stopCh <-chan
 	flags.DurationVar(&o.DiscoveryInterval, "discovery-interval", o.DiscoveryInterval, ""+
 		"interval at which to refresh API discovery information")
 	flags.StringVar(&o.PrometheusURL, "prometheus-url", o.PrometheusURL,
-		"URL and configuration for connecting to Prometheus.  Query parameters are used to configure the connection")
+		"URL for connecting to Prometheus.  Query parameters are used to configure the connection")
+	flags.BoolVar(&o.PrometheusAuthInCluster, "prometheus-auth-incluster", o.PrometheusAuthInCluster,
+		"use auth details from the in-cluster kubeconfig when connecting to prometheus.")
+	flags.StringVar(&o.PrometheusAuthConf, "prometheus-auth-config", o.PrometheusAuthConf,
+		"kubeconfig file used to configure auth when connecting to Prometheus.")
 	flags.StringVar(&o.LabelPrefix, "label-prefix", o.LabelPrefix,
 		"Prefix to expect on labels referring to pod resources.  For example, if the prefix is "+
 			"'kube_', any series with the 'kube_pod' label would be considered a pod metric")
-
 	return cmd
+}
+
+// makeHTTPClient constructs an HTTP for connecting with the given auth options.
+func makeHTTPClient(inClusterAuth bool, kubeConfigPath string) (*http.Client, error) {
+	// make sure we're not trying to use two different sources of auth
+	if inClusterAuth && kubeConfigPath != "" {
+		return nil, fmt.Errorf("may not use both in-cluster auth and an explicit kubeconfig at the same time")
+	}
+
+	// return the default client if we're using no auth
+	if !inClusterAuth && kubeConfigPath == "" {
+		return http.DefaultClient, nil
+	}
+
+	var authConf *rest.Config
+	if kubeConfigPath != "" {
+		var err error
+		loadingRules := &clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeConfigPath}
+		loader := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
+		authConf, err = loader.ClientConfig()
+		if err != nil {
+			return nil, fmt.Errorf("unable to construct  auth configuration from %q for connecting to Prometheus: %v", kubeConfigPath, err)
+		}
+	} else {
+		var err error
+		authConf, err = rest.InClusterConfig()
+		if err != nil {
+			return nil, fmt.Errorf("unable to construct in-cluster auth configuration for connecting to Prometheus: %v", err)
+		}
+	}
+	tr, err := rest.TransportFor(authConf)
+	if err != nil {
+		return nil, fmt.Errorf("unable to construct client transport for connecting to Prometheus: %v", err)
+	}
+	return &http.Client{Transport: tr}, nil
 }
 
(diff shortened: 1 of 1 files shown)


huxiaoliang:
@DirectXMan12  @aschepis  @brancz  @luxas  could you help review this patch? Thanks in advance!

tcolgate:
The only part of this that really makes sense is the optional
tlsInsecureSkipVerify. http.DefaultClient can talk SSL, and will use the
hosts CA to verify the server. Including the remote hosts TLS key is
incorrect, as is including its certificate.

huxiaoliang:
@tcolgate  thanks for your comments, but k8s-prometheus-adapter should have a way (prometheus-cacert, prometheus-cert, prometheus-key etc) to specify the client side key and cert so that talk to remote prometheus server, any comments?

tcolgate:
None of those are needed to talk to a HTTPS enabled Prometheus. They might
be needed if the Prometheus server is doing HTTPS client authentication.
That may be worth supporting, but I suspect basic authentication over HTTPS
would be the more common case.

huxiaoliang:
@tcolgate  It is yes, that my Prometheus server behind a nginx which is enabled authentication according to [official guideline](https://prometheus.io/docs/introduction/faq/#why-don't-the-prometheus-server-components-support-tls-or-authentication?-can-i-add-those?),  my patch will support this case, it is worth supporting.

tcolgate:
I am not sure what official guide you are referring to. The links you refer
to demonstrate using nginx to provide HTTP Basic auth, not TLS client
certificate authentication.
 To be clear, the existing code does support access a HTTPS enabled
Prometheus, as long as the default CA validates the servers certificate.
  The part of your change that allows TLSInsecureSkipVerify  worth
including.
  Personally I wouldnt bother with the TLS client certificate support, I
don't think that is a very common use case.Some support for basic auth,
and/or bearer tokens is probably worth while.
  I'm not one of the maintainers though, so not my call.

tcolgate:
The guide you are linking to only specifies HTTP Basic auth, not SSL. If
you have changed this configuration to require TLS Client certificate
verification, then you will need the changes you have suggested, but that
would be a rare configuration.
The most likely cause of problems with using HTTPS is the verification of
the remote servers cert. In order to talk to a remote https prometheus,
k8s-prometheus-adapter would need to be able to trust the remote server's
cert using the name that it is accessing it. This would
require /etc/ssl/certs/ca-certificates.crt in the k8s-prometheus-adapter to
include the CA used to sign the cert. If the cert is signed by the cluster
then you can either arrange for the CA cert to be in a config map, or you
can symlink in the copy from the pods serviceaccount. link
/etc/ssl/certs/ca-certificates.crt to
/var/run/secrets/kubernetes.io/serviceaccount/ca.crt

huxiaoliang:
@tcolgate  Yes, I need both server and client certificate verification, so my enhancement is worth to have.  Regarding to your steps,  It seems that not a k8s style authentication by ssl/tls (depends on system ca bundle, the path is different base on Linux release), I just follow [k8s api server client certificate authentication](https://github.com/kubernetes-incubator/apiserver-builder/blob/master/docs/concepts/auth.md#client-certificate-authentication)  to address my case. Any way, thanks for you discuss with me on this topic.

DirectXMan12:
So, I'd prefer here to have something like a `kubeconfig` for connecting to the remote prometheus -- different people use different solutions.  Some use SSL (like above), others use kubernetes-native token proxies (like in OpenShift), and having an option for each would be a lot of options.

brancz:
I think it would be viable to have that as part of the general configuration of the adapter. (as in what is proposed in #46 )

huxiaoliang:
@DirectXMan12 I updated codes according to your comments, below is my patch summarize:

1.  introduce 2 new cli parameters for connecting prometheus according to different user case

    --prometheus-use-auth 
     default value is false, that means adapter connects to prometheus which is not enable auth, true means adapter will do auth with prometheus base on `InClusterConfig` or customize `kubeconfig` style config file.

    --prometheus-auth-config
      adapter use `kubeconfig` style config file to initialize http client  base on k8s functionality, this client will support token, ssl/tls, and username/password auth with server

2. test case:

 1)  old case (no auth enable on prometheus)
```
/adapter --secure-port=7443 --tls-cert-file=/var/run/serving-cert/serving.crt --tls-private-key-file=/var/run/serving-cert/serving.key --logtostderr=true --prometheus-url=http://10.1.120.206:9090 --prometheus-use-auth=false  --metrics-relist-interval=30s --rate-interval=5m --v=10 
```      
 2) ssl/tls auth with prometheus  (this is my real case)
  
```
/adapter --secure-port=6443 --tls-cert-file=/var/run/serving-cert/serving.crt --tls-private-key-file=/var/run/serving-cert/serving.key --logtostderr=true --prometheus-url=https://monitoring-prometheus:9090 --prometheus-use-auth=true --prometheus-auth-config=/conf --metrics-relist-interval=30s --rate-interval=5m --v=10
```
conf file

```
apiVersion: v1
clusters:
- cluster:
    insecure-skip-tls-verify: true
    server: https://monitoring-prometheus:9090
  name: mycluster.icp
contexts:
- context:
    cluster: mycluster.icp
    namespace: default
    user: test
  name: mycluster.icp-context
current-context: mycluster.icp-context
kind: Config
preferences: {}
users:
- name: test
  user:
    client-certificate: /opt/test/monitoring/certs/tls.crt
    client-key: /opt/test/monitoring/certs/tls.key
```
BTW: due to issue [how-do-i-make-an-https-call-in-a-busybox-docker-container-running-go ](https://stackoverflow.com/questions/45388934/how-do-i-make-an-https-call-in-a-busybox-docker-container-running-go), I have to update Dockerfile as well

huxiaoliang:
@DirectXMan12  I revert the change on Dockerfile and handle it outside of adapter (your comments from slack),  for example below ticket:
https://github.com/kubernetes/heapster/issues/1611, could you help review

Issue #54, opened 2018-03-14
miss LICENSE file


huxiaoliang:
The best practice that the repository should have `LICENSE `file according to [here ](https://help.github.com/articles/licensing-a-repository/ )so that tells others what they can and can't do with the source code,  it's important to make an informed decision.  

I notice that many files tag `Apache License, Version 2.0` license, but we should make it more clear by `LICENSE `file

DirectXMan12:
closed by #55

PR #55, opened 2018-03-15
Fix license issue

huxiaoliang:
fix issue #54


huxiaoliang:
hi @DirectXMan12  I create this PR to address License issue, but I don't know  whether is it what you want, please help review, if not ,could you show me the reason?  Thanks in advance.

PR #93, opened 2018-07-22
sets the right path for config file

aabed:
The configmap is mounted under /etc/adapter/config.yaml instead of /default/adapter-config.yaml


DirectXMan12:
whoops, good catch, must have been a last-minute typo.  Thanks!

russiancook:
@aabed Thanks for the fix been going crazy thinking I configured something wrong

PR #94, opened 2018-07-23
Use type kubernetes.io/tls for adapter TLS certificates

zioproto:
This PR uses standard `kubernetes.io/tls` secret type for the adapter TLS certificates.

Also I documented a little bit in detail how the user can get this certificates ready.

I fixed the deployment so that it works. The config file reference was wrong. `/etc/adapter/config.yaml` works for me.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -2,13 +2,15 @@ Example Deployment
 ==================
 
 1. Make sure you've built the included Dockerfile with `make docker-build`. The image should be tagged as `directxman12/k8s-prometheus-adapter:latest`.
-   
+
 2. Create a secret called `cm-adapter-serving-certs` with two values:
    `serving.crt` and `serving.key`. These are the serving certificates used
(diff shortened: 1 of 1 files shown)


zioproto:
@DirectXMan12 fixing `/etc/adapter/config.yaml` in this PR is a bit offtopic, let me know if you want a separate PR for that.

DirectXMan12:
IMO, it's actually a bad practice to do this -- that CA is also used for cluster bootstrapping, so the cert is valid as a client cert.  We should probably have a separate CA for serving certs...

RE: the config path, just got another PR from someone else to fix that.  It's merged now.

zioproto:
You are right. I found that the kube-prometheus has scripts to actually use a new CA for that. I reworked the patch just linking those scripts to the readme. Thanks

DirectXMan12:
whoops, that title is wrong.  oh well

PR #95, opened 2018-08-07
[docs] Config Walkthroughs and FAQs

DirectXMan12:
This adds a more hands-on configuration example, plus a few FAQs.


DirectXMan12:
Thanks to @aabed and atomoras for rightly pointing out that configuration
and labeling needed a bit more documentation!

atomaras:
I like it! 
One suggestion would be to give a more prominent position to the HPA in the walkthrough. 
As a user this is my initial viewpoint and my ultimate goal. 
_What should my HPA yaml look like for it to actually work with custom metrics?_
Then I would say that this HPA yaml will translate to hpa controller making this rest call to figure out the metric i.e `$KUBERNETES/apis/custom.metrics.k8s.io/v1beta1/namespaces/production/pods/*/http_requests_per_second`

Then I would jump into how custom metrics can be configured to expose this exact URL for HPA to be happy.
This HPA intro (goal first approach) will give readers all the context they need to be able to follow along and understand why we do all these complex configurations.

README.md
```diff
@@ -76,3 +83,75 @@ attention to:
...
+information.
+
+### Why isn't my metric showing up?
+
+First, check your configuration.  Does it select your metric?  You can find the
+[default configuration](/deploy/custom-metrics-config-map.yaml) in the deploy directory,
+and more information about configuring the adapter in the [docs](/docs/config.md).
+
+Next, check if the discovery information looks right.  You should see the metrics showing
+up as associated with the resources you expect at. `/apis/custom.metrics.k8s.io/v1beta1/`.
```
tanner-bruce: This could maybe be `kubectl get --raw /apis/custom.metrics.k8s.io/v1beta1 | jq .`

docs/config-walkthrough.md
```diff
@@ -0,0 +1,222 @@
...
+`kubernetes_pod_name` represents the pod name.  Since these names don't
+quite follow a consistent pattern, we use the `overrides` section of the
+`resources` field in our rule:
+
+```yaml
+rules:
+- seriesQuery: 'http_requests_total{kubernetes_namespace!="",kubernetes_pod_name!=""}'
+  resources:
+    overrides:
+      kubernetes_namespace: {resource: "namespace"}
```
tanner-bruce: Is there somewhere that lists the available resource types? It might be good to include a link to that around this section.

docs/config-walkthrough.md
```diff
@@ -0,0 +1,222 @@
...
+      kubernetes_pod_name: {resource: "pod"}
+```
+
+This says that each label represents its corresponding reosurce. Since the
```
tanner-bruce: typo: reosurce

docs/config-walkthrough.md
```diff
@@ -0,0 +1,222 @@
...
+    as: "${1}_per_second"
+  metricsQuery: 'sum(rate(<<.Series>>{<<.LabelMatchers>>}[2m])) by (<<.GroupBy>>)'
+```
+
+The adapter will automatically fill in the right series name, label
+matchers, and group-by clause, depending on what we put into the API.
+Since we're only working with a single metric anyway, we could replace
+`<<.Series>>` with `http_requests_total`.
+
+Now, if we put run an instance of the Prometheus adapter with this
```
tanner-bruce: if we `put run`

docs/walkthrough.md
```diff
@@ -146,7 +146,10 @@ You may also need to modify the ConfigMap containing the metrics discovery
 configuration.  If you're using the Prometheus configuration described
 above, it should work out of the box in common cases.  Otherwise, read the
 [configuration documentation](/docs/config.md) to learn how to configure
-the adapter for your particular metrics and labels.
+the adapter for your particular metrics and labels.  The [configuration
+walkthrough](/docs/config-walkthrough.md) gives an end-to-end
+configuration tutorial for configure the adapter for a scenario similar to
```
tanner-bruce: for configuring

docs/walkthrough.md
```diff
@@ -270,6 +273,20 @@ Try fetching the metrics again.  You should see an increase in the rate
 after the collection interval specified in your Prometheus configuration
 has elapsed.  If you leave it for a bit, the rate will go back down again.
 
+### Quantity Values
+
+Notice that the API uses Kubernetes-style quantities to describe metric
+values.  These quantities use SI suffixes instead of decimal points.  The
```
tanner-bruce: There are two spaces after some periods here

README.md
```diff
@@ -76,3 +83,75 @@ attention to:
...
+You're probably switching between whole numbers (e.g. `10`) and milli-quantities (e.g. `10500m`).
+Worry not!  This is just how Kubernetes represents fractional values.  See the
+[Quantity Values](/docs/walkthrough.md#quantity-values) section of the walkthrough for a bit more
+information.
+
+### Why isn't my metric showing up?
+
+First, check your configuration.  Does it select your metric?  You can find the
+[default configuration](/deploy/custom-metrics-config-map.yaml) in the deploy directory,
+and more information about configuring the adapter in the [docs](/docs/config.md).
```
discordianfish: It would be helpful to have a command to return all metrics. Not only the registered metric names but the actual per pod/object metrics. That would help to validate this.

README.md
discordianfish: Agree, at least it should mention `kubectl get --raw` somewhere here.

discordianfish:
Great, that will help a lot!

I'd also add some words around using object vs pod metrics. Another thing I haven't figured out (not needing it right now) is how to make the HPA scale *down* of a metric goes *up*.

README.md
```diff
@@ -76,3 +83,75 @@ attention to:
...
+
+### My query contains multiple metrics, how do I make that work?
+
+It's actually fairly straightforward, if a bit non-obvious.  Simply choose one
+metric to act as the "discovery" and "naming" metric, and use that to configure
+the "discovery" and "naming" parts of the configuration.  Then, you can write
+whichever metrics you want in the `metricsQuery`!  The series query can contain
+whichever metrics you want, as long as they have the right set of labels.
+
+For example, if you have two metrics `foo_total` and `foo_count`, you might write
```
tanner-bruce: It could be useful for clarity to say that `foo_total` has the label `system_name` which represents the actual `node`

DirectXMan12:
TODO: for me: make sure to add an FAQ entry about flickering metrics

README.md
itskingori: I agree. I'm currently struggling with the config and was hoping there's an easy way to list what it can "see".

itskingori:
@DirectXMan12 This is really great, make the configuration clearer.

README.md
discordianfish: This only shows what metrics it knows about but you still might not have a metric with the right labels for a particular HPA. In that case it would be useful to have a expression that also returns the per-object/pod metrics.

itskingori: Yeah, initially I came to the same conclusion eventually. 😅

docs/config-walkthrough.md
DirectXMan12: There is a command, but it can change at any time due to CRDs, aggregated API servers, etc.  I'll add a note about it.

README.md
DirectXMan12: can you file a separate issue for that?

Issue #90, opened 2018-07-10
Images and a chart


plumdog:
Now that I've got k8s-prometheus-adapter up and working, I am utterly convinced of its necessity.

However, getting it working wasn't straightforward as the linked example is a bit old, and I was lacking in some prometheus and kubernetes-investigation skills that I have learned along the way.

I think two things would help to make k8s-prometheus-adapter easier to get going with:

- Official images built and made available on quay/dockerhub/other.
- A helm chart added to https://github.com/kubernetes/charts

This would give any docs/walkthroughs something to work from (eg, "this walkthrough uses chart version x and image version y") to reduce any version-related snafus.

Given that I think I have the basis of a working chart (I started with https://github.com/madeden/charts/tree/master/custom-metrics-apiserver and have tinkered), I think I can be useful there. Also, I'm keen to document my findings somewhere useful, probably by adding bits to the walkthrough. (In particular, how to get this working with metrics from the nginx-ingress-controller).

However, without images, the above becomes rather tricky. I notice that https://hub.docker.com/r/directxman12/k8s-prometheus-adapter-amd64/tags/ does have a quite recent "latest", but nothing else. Even just tagging with the Git SHA would be super useful.

Once those images exist, I'd been keen to contribute a chart to the charts repo, and update the walkthrough accordingly (for anyone using helm).

DirectXMan12:
There are official images built by the CI (`directxman12/k8s-prometheus-adapter-${ARCH}`), but there was a hiccup in that (as you noticed) due to an issue with what appeared to be github rate limiting.  I'll tag another release to trigger another tag.

I'd be happy to have an official chart submitted, let me know if you submit a PR.  I'd also *love* any contributions to the walkthrough :-).

steven-sheehy:
I've submitted a Helm chart to to charts/incubator: https://github.com/helm/charts/pull/7159

steven-sheehy:
@DirectXMan12 I think this issue can be closed now. prometheus-adapter is now in the official chart repository under [stable/prometheus-adapter](https://github.com/helm/charts/tree/master/stable/prometheus-adapter).

plumdog:
Amazing stuff.

Yup, I agree this can now be closed.

Perhaps worth adding a link to that in the docs somewhere, but I guess this issue sort of achieves that anyway.

PR #98, opened 2018-08-15
Add docs on helm installation

tomkerkhove:
Add docs on helm installation in README.

Relates to #12 #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,13 @@ metrics API
 suitable for use with the autoscaling/v2 Horizontal Pod Autoscaler in
 Kubernetes 1.6+.
 
+Quick Links
+-----------
+
+- [Config walkthrough](docs/config-walkthrough.md) and [config reference](docs/config.md).
+- [End-to-end walkthrough](docs/walkthrough.md)
+- [Deployment info and files](deploy/README.md)
+
 Installation
 -------------
 If you're a helm user, a helm chart is listed on the Kubeapps Hub as [stable/prometheus-adapter](https://github.com/helm/charts/blob/master/stable/prometheus-adapter/README.md).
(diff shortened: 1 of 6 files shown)


DirectXMan12:
So, I'd prefer to just have a link to the helm chart -- I don't necessarily want people to be filing bugs about the helm chart here, since it'll be hard for me to support them.  I'm happy to link to it, but it's not the "official" supported installation method for the moment (the official supported installation method is `kubectl create -f deploy/manifests`).

However, I'm open to count-arguments on this matter :-).

DirectXMan12:
Yeah, that link is perfect.  Maybe say something like

> If you're a helm user, a helm chart is listed on the Kubeapps Hub as [stable/prometheus-adapter](https://hub.kubeapps.com/charts/stable/prometheus-adapter)

or something (personally, I'm not a huge helm fan, but I understand why people like it :-) )

tomkerkhove:
Sounds good to me! Do you prefer to have only that one-liner or just adapt current phrasing?

Ha, that's interesting - Probably because you prefer to know the details/advanced stuff?

DirectXMan12:
> Probably because you prefer to know the details/advanced stuff?

Yeah, a bit.  There's a few reasons -- I'm not a huge fan of the tiller architecture, but beyond that, I think it's at least partially the wrong abstraction, at least for me.  It's great if you only have one or two knobs, but once you have a bunch of knobs, you end up with a second complicated layer on top of the kubernetes YAML that mirrors a lot of the Kubernetes YAML, only slightly differently.  It also means that you don't get used to how Kubernetes objects actually look, which means that when you go to debug your cluster, you've got a completely different "language" (kubernetes YAML) than the what you normally use (per-chart helm yaml), and you've got less of an idea as to what's actually going on.  Furthermore, if you already know Kubernetes, you have to re-learn where particular knobs are in any given Helm chart.

I very much prefer the approach [kustomize](https://github.com/kubernetes-sigs/kustomize) took -- you still work with kubernetes objects, and customization is done by layering patches on top, plus some basic primatives for pulling files into secrets and configmaps.  This means that you still interact with snippets of Kubernetes YAML (to make the patches), and you can customize any knob in the underlying Kubernetes object, but because they're patches, they're still relatively small and easy to read.

That's not to say that I don't sympathize with what Helm is trying to do -- customizing Kubernetes objects, and packing that for deployment, can be crazy complicated for a newcomer.

DirectXMan12:
>  Do you prefer to have only that one-liner or just adapt current phrasing?

The one liner is fine, but if you've got better wording, feel free to use that -- that was just off the top of my head.

tomkerkhove:
@DirectXMan12 I've updated it and think this looks ok. Feel free to let me know if you agree.

tomkerkhove:
Thanks for considering, reviewing and merging this PR8

PR #102, opened 2018-08-20
Reduce metric namer label-GVR logspam

### LINKED ISSUE #101: Prints lots of `unable to normalize group-resource le from label` errors

DirectXMan12:
It's fairly common to have a label pattern that matches lots of
irrelevant labels, so this makes the "error" about being unable to
normalize/resolve a label to a GVR a V(9) info log, instead of an error
log.

Closes #101


DirectXMan12:
cc @discordianfish

PR #103, opened 2018-08-20
Remove extraneous `/adapter` from deployment args

DirectXMan12:
There was an extraneous `/adapter` in the deployment args that was
causing people issues.  This removes it.


DirectXMan12:
cc @tanner-bruce

tanner-bruce:
LGTM.

PR #106, opened 2018-08-28
Resource Metrics API Support

DirectXMan12:
This PR introduces support for serving the resource metrics API.  It uses a new config section to allow configuring the individual queries used for container and node metrics, and allows reporting accurate window sizes.


DirectXMan12:
cc @brancz

pkg/naming/resource_converter.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (r *resourceConverter) ResourcesForSeries(series prom.Series) ([]schema.GroupResource, bool) {
+	// use an updates map to avoid having to drop the read lock to update the cache
+	// until the end.  Since we'll probably have few updates after the first run,
+	// this should mean that we rarely have to hold the write lock.
+	var resources []schema.GroupResource
+	updates := make(map[pmodel.LabelName]schema.GroupResource)
+	namespaced := false
+
+	// use an anon func to get the right defer behavior
```
brancz: there are no returns in this anonymous function so we can just lock in the beginning and unlock afterwards

pkg/naming/resource_converter.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+		}
+	}()
+
+	// update the cache for next time.  This should only be called by discovery,
+	// so we don't really have to worry about the gap between read and write locks
+	// (plus, we don't care if someone else updates the cache first, since the results
+	// are necessarily the same, so at most we've done extra work).
+	if len(updates) > 0 {
+		r.labelResourceMu.Lock()
```
brancz: same here, we can just lock before and unlock after the `for` statement

pkg/resourceprovider/provider.go
```diff
@@ -0,0 +1,343 @@
...
+	cpu, mem queryResults
+	err error
+}
+
+func (p *resourceProvider) GetContainerMetrics(pods ...apitypes.NamespacedName) ([]provider.TimeInfo, [][]metrics.ContainerMetrics, error) {
+	if len(pods) == 0 {
+		return nil, nil, fmt.Errorf("no pods to fetch metrics for")
+	}
+
+	// TODO(directxman12): figure out how well this scales if we go to list 1000+ pods
```
brancz: yes we should make sure this goes along with openshift scalability requirements

cmd/config-gen/utils/default.go
```diff
@@ -89,5 +89,25 @@ func DefaultConfig(rateInterval time.Duration, labelPrefix string) *MetricsDisco
 				MetricsQuery: fmt.Sprintf("sum(rate(<<.Series>>{<<.LabelMatchers>>}[%s])) by (<<.GroupBy>>)", pmodel.Duration(rateInterval).String()),
 			},
 		},
+
+		ResourceRules: &ResourceRules{
+			CPU: ResourceRule{
+				ContainerQuery: fmt.Sprintf("sum(rate(container_cpu_usage_seconds_total{<<.LabelMatchers>>}[%s])) by (<<.GroupBy>>)", pmodel.Duration(rateInterval).String()),
```
brancz: I'm curious about this, do you think for "node" metrics, should we sum up the container usage or actually use the node-exporter's metrics. The latter would probably be the more correct one I would say, but induce an additional dependency.

DirectXMan12: I'm matching what the summary API does right now, which appears to be looking at the root cgroup.

DirectXMan12: (which I do think makes some sense)

pkg/naming/resource_converter.go
DirectXMan12: defer avoids issues if a panic occurs

DirectXMan12:
@brancz I think this is no longer WIP, PTAL

DirectXMan12:
(I've also added in the ability to disable either API and equivalent functionality by simply not specifying rules for it)

PR #136, opened 2018-11-20
Add a separate flag for 'start' parameter

### LINKED ISSUE #135: `--metrics-relist-interval` flag should be decoupled from "start" parameter
Currently `--metrics-relist-interval` not only affects how often the list of metrics is polled, but also is used for "start" parameter sent to Prometheus: https://github.com/DirectXMan12/k8s-prometheus-adapter/blob/3bd75f5c3ae58b6135d3d10e17aa8bf99acc8170/pkg/custom-provider/provider.go#L215

The issue was previously reported in #64, but the proper fix is to introduce a separate parameter for determining "start", e.g. `--metrics-expire-period`.

I want to keep the `--metrics-relist-interval` as short as 30 seconds to see new metrics as soon as they appear, but don't want the flakiness that was previously reported in #64.

nilebox:
Fixes #135


cmd/adapter/adapter.go
```diff
@@ -53,6 +53,8 @@ type PrometheusAdapter struct {
 	AdapterConfigFile string
 	// MetricsRelistInterval is the interval at which to relist the set of available metrics
 	MetricsRelistInterval time.Duration
+	// MetricsStartDuration is the period to query available metrics for
+	MetricsStartDuration time.Duration
```
DirectXMan12: nit: this name is not immediately obvious.  Perhaps `max-metrics-age` or something

cmd/adapter/adapter.go
```diff
@@ -83,6 +85,8 @@ func (cmd *PrometheusAdapter) addFlags() {
 			"and custom metrics API resources")
 	cmd.Flags().DurationVar(&cmd.MetricsRelistInterval, "metrics-relist-interval", cmd.MetricsRelistInterval, ""+
 		"interval at which to re-list the set of all available metrics from Prometheus")
+	cmd.Flags().DurationVar(&cmd.MetricsStartDuration, "metrics-start-duration", cmd.MetricsStartDuration, ""+
```
DirectXMan12: can we default this to something sane based the `metrics-relist-interval` so that we don't confuse people  who are expecting the existing behavior?  e.g. if someone sets `metrics-relist-interval` to 20 minutes, they probably don't expect the max metrics age to be 10 minutes, and if someone sets it to 30 seconds, they probably also don't expect the age to be 10 minutes.  Maybe `2 * relist interval` by default or something?

nilebox:
@DirectXMan12 done, please review.

cmd/adapter/adapter.go
nilebox: Agreed that I can make it longer than `metrics-relist-interval` by default, i.e. `20m`. That would suggest that max age should be at least twice as high as relist interval.

Why setting max age to `10m` with relist interval `30s` is bad though? I think it's perfectly fine actually :) e.g. even if Prometheus is configured to be super slow and refreshes metrics once every `10m`, I still don't want to have a potential delay of `9m:59s` once a new metric is discovered by Prometheus, so I will set it to `30s`.

cmd/adapter/adapter.go
nilebox: Renamed to `metrics-max-age` to be consistent with existing `metrics-relist-interval` flag name.

cmd/adapter/adapter.go
nilebox: @DirectXMan12 I can probably add a validation code that would require the `max-age >= 2 * relist-interval`?

DirectXMan12: no validation needed.  My main concern was that it means metrics that are no longer valid will show up in discovery for a lot longer.  If you have things coming and going quickly, this could lead to a lot more metrics to process during discovery than you intend.  It's not a big deal here, though, and we can always follow up later.

DirectXMan12:
LGTM :+1:

PR #141, opened 2018-12-03
Support custom ca certificate to talk to Prometheus

metalmatze:
We need to be able to use another certificate to talk to Prometheus than talking to Kubernetes.
Therefore this PR adds another HTTP client that gets a certificate loaded from a ca file.

/cc @DirectXMan12 @brancz @s-urbaniak

### DIFF
--- a/cmd/adapter/adapter.go
+++ b/cmd/adapter/adapter.go
@@ -76,14 +76,14 @@ func (cmd *PrometheusAdapter) makePromClient() (prom.Client, error) {
 			return nil, err
 		}
 		httpClient = prometheusCAClient
-		fmt.Println("successfully loaded ca file")
+		glog.Info("successfully loaded ca from file")
 	} else {
 		kubeconfigHTTPClient, err := makeKubeconfigHTTPClient(cmd.PrometheusAuthInCluster, cmd.PrometheusAuthConf)
 		if err != nil {
 			return nil, err
 		}
 		httpClient = kubeconfigHTTPClient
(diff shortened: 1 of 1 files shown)


cmd/adapter/adapter.go
```diff
@@ -246,3 +270,25 @@ func makeHTTPClient(inClusterAuth bool, kubeConfigPath string) (*http.Client, er
 	}
 	return &http.Client{Transport: tr}, nil
 }
+
+func makePrometheusCAClient(caFilename string) (*http.Client, error) {
+	pool, err := x509.SystemCertPool()
```
brancz: I think [`NewCertPool()`](https://github.com/golang/go/blob/41e62b8c49d21659b48a95216e3062032285250f/src/crypto/x509/cert_pool.go#L21), would suffice here and wouldn't require having a cert in place (which busybox doesn't).

metalmatze:
I've just tested the updated with `NewCertPool` on my OpenShift cluster. Works!

brancz:
lgtm :+1: thanks!

brancz:
While that may be theoretically possible, it's rather strange to configure access to Prometheus with a "kubeconfig", especially if parts of the kubeconfig are ignored and instead are taking from the flag. I do have to admit though that I didn't realize this is what it did. I read kubeconfig and immediately associated it with communicating with Kubernetes.

metalmatze:
I agree with @brancz. I have looked at all the flags and didn't see this option... :confused:

PR #143, opened 2018-12-05
cmd/adapter: add prometheus bearer token configuration

s-urbaniak:
Often prometheus is gated by some proxy requiring an auth bearer
token. Currently there is no possibility to configure one except for
providing a full-fledged kubeconfig.

This fixes it by adding a new flag pointing to an optional file
containing the auth bearer for prometheus communication.


brancz:
lgtm :+1: 

(as a side note for next time, we really should have done the full integration, before cutting a release, now we need to do v0.5.0 .. nothing bad, just a side note)

PR #148, opened 2018-12-11
Fix empty pod or node list resulting in an error

### LINKED ISSUE #147: resource metrics: no pods to fetch metrics for
A cluster that has a namespace with no pods where a request for said namespace is done returns an error along the lines of:

```
Error from server: Error while fetching pod metrics for selector  in namespace "default": no pods to fetch metrics for
```

This is done here: https://github.com/DirectXMan12/k8s-prometheus-adapter/blob/7a421bb91e9ea52be4e26db22392c36c8af38bc8/pkg/resourceprovider/provider.go#L122-L124

My feeling is that this shouldn't error, but just return an empty list, and looking at what the [metrics-server does](https://github.com/kubernetes-incubator/metrics-server/blob/ca29d12df6c984c2ed3816772ef92fbb0be16fe5/pkg/provider/sink/sinkprov.go#L80-L119), it seems that that's what is intended.

@DirectXMan12 @squat @s-urbaniak @metalmatze

brancz:
Fixes #147 

Tested this on my cluster and works as expected. Technically I don't see how the node list should ever be empty, but I fixed the case anyways. Also let me know if you prefer nil or empty lists rather, I have no strong opinion on either.

@DirectXMan12 @s-urbaniak @squat @metalmatze

### DIFF
--- a/pkg/resourceprovider/provider.go
+++ b/pkg/resourceprovider/provider.go
@@ -120,7 +120,7 @@ type nsQueryResults struct {
 
 func (p *resourceProvider) GetContainerMetrics(pods ...apitypes.NamespacedName) ([]provider.TimeInfo, [][]metrics.ContainerMetrics, error) {
 	if len(pods) == 0 {
-		return nil, nil, fmt.Errorf("no pods to fetch metrics for")
+		return nil, nil, nil
 	}
 
 	// TODO(directxman12): figure out how well this scales if we go to list 1000+ pods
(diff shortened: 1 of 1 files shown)


metalmatze:
I don't have a strong opinion on this either. 
Maybe it's better for the caller to return with empty list and not `nil` three times.

squat:
This looks good to me as is :)

s-urbaniak:
If the caller asserts `len([]provider.TimeInfo) == 0` and `len([][]metrics.ContainerMetrics) == 0`, then both will be true if `return nil, nil, nil` is being used, that's quite idiomatic here.

I am :+1: on the change.

PR #158, opened 2019-01-09
Edited Makefile to add cross build support for s390x. 

guirish:
As golang:1.10 has now multi-arch support added for s390x removing explicit GOIMAGE specification in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,6 @@ ifeq ($(ARCH),ppc64le)
 endif
 ifeq ($(ARCH),s390x)
 	BASEIMAGE?=s390x/busybox
-	GOIMAGE=s390x/golang:1.10
 endif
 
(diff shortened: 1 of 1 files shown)


guirish:
Team any updates on the PR merger?

guirish:
Earlier when [support ](https://github.com/DirectXMan12/k8s-prometheus-adapter/commit/5e307fb2b45e535a9a88b16cc0e7be0c62d6c5b8) was added for s390x GOIMAGE referred to golang:1.8. It was an multi-arch image but did not have s390x support. But now as GOIMAGE has been updated to golang:1.10 it has multi-arch support added for [s390x](https://github.com/docker-library/golang/blob/master/1.10/stretch/Dockerfile) too. I have locally cross compiled the binaries with above changes.

Issue #169, opened 2019-02-25
Disable TLS verification for prometheus


sebastianneb:
Hey there,

I am currently facing the issue, that our prometheus service is secured via https via some url e.g.`prometheus.test.de`. Within the config I set the url to cluster-internal svc url like `https://prometheus.monitoring.svc.cluster.local`. I also mount the CA from letsencrypt via `--prometheus-ca-file=/etc/prom/certs/cert.pem` but still (as expected) the tls validation fails due to `certificate is valid for *.test.de, not prometheus.monitoring.svc.cluster.local`.

As I looked in the docs and was not able to find any flag there, is it possible to disable this tls verification when connecting to prometheus (and I know, someone should not do this, but still)?

sebastianneb:
So I added this PR https://github.com/DirectXMan12/k8s-prometheus-adapter/pull/170, hoping that this will enable what I try to achieve. Thoughts welcome of course.

s-urbaniak:
@sebastianneb generally, I'd recommend not to disable TLS verification in this case, especially if you are running the prometheus service in production.

I'd rather recommend you to set the internal cluster DNS name `prometheus.monitoring.svc.cluster.local` as a SAN (subject alternative name) in the certificate request towards acme.

PR #170, opened 2019-02-26
Skip TLS verification for Issue #169

schostin:
This MR should hopefully add the functionality issued in #169.

Since I have no experience in go, I have no idea If what i did works or if tests / etc. might be missing. So improvements / comments are very welcome here!


sebastianneb:
So you mean some e.g. ConfigMap / file to hold all properties for `tls.config`?

s-urbaniak:
@sebastianneb prometheus adapter has a config struct, read via the `--prometheus-auth-config` parameter. I agree, the skip tls settings could be added here https://github.com/DirectXMan12/k8s-prometheus-adapter/blob/master/pkg/config/config.go.

Issue #119, opened 2018-10-18
AWS EKS: cluster doesn't provide client-ca-file


toriacht:
Hi,

I am trying to deploy on AWS EKS which now supports Kubernetes HPA [AWS Blog](https://aws.amazon.com/blogs/opensource/horizontal-pod-autoscaling-eks/). I successfulyy deployed Prometheus via helm but now when i try to deploy the adapter i get the following error

`kubectl logs promadapter-prometheus-adapter-6d68c79bbc-pjgvw
I1018 11:41:13.278829       1 serving.go:273] Generated self-signed cert (/tmp/cert/apiserver.crt, /tmp/cert/apiserver.key)
Error: cluster doesn't provide client-ca-file
...
...
panic: cluster doesn't provide client-ca-file

goroutine 1 [running]:
main.main()
        /go/src/github.com/directxman12/k8s-prometheus-adapter/cmd/adapter/adapter.go:41 +0x13b
`

How can i point it towards this ca file?
I tried pointing to .kube config but this did nothing

`helm install --name custommetrics stable/prometheus-adapter --set --lister-kubeconfig=/root/.kube/config`

Any tips much appreciated...

bhavin192:
Duplicate of #108

bhavin192:
Hey, this was fixed in #110, can you please try image with `latest` tag? 
Or you can also image mentioned in https://github.com/DirectXMan12/k8s-prometheus-adapter/issues/108#issuecomment-424817185

twittyc:
So I tried with the latest built image, but I'm running into issues authenticating against the apiserver. Do I need to be providing the `--requestheader-client-ca-file` flag? I've tried with and without it with no luck. @bhavin192 Do you have an example of the config you used with your deployment? Thanks!

bwang221:
I ran into this today. any update?

bhavin192:
@twittyc I'm using the helm chart to install Prometheus adapter, and I'm using image with tag `pr110` as mentioned here https://github.com/DirectXMan12/k8s-prometheus-adapter/issues/108#issuecomment-424817185

VinayVanama:
> Hey, this was fixed in #110, can you please try image with `latest` tag?
> Or you can also try image mentioned in [#108 (comment)](https://github.com/DirectXMan12/k8s-prometheus-adapter/issues/108#issuecomment-424817185)

Hi ! Even I'm facing with the same issue, where will be my ca.crt file located ?

[...]

PR #172, opened 2019-03-04
Add Mozilla CA cert store in the docker image and enable building from macos

nicolaevladescu:
Hi.

The initial reason of this PR has to do with us using a valid `*.random.com` certificate for our ssl enabled prometheus server, inside AWS EKS. And getting this error `x509: failed to load system roots and no roots provided`.

While looking at the source code i found the undocumented (and not available in the official stable chart) option `--prometheus-ca-file` that we could have used to mitigate this, but i decided to open this PR anyway, maybe somebody else could use the features that it introduces.

Related issue: #119

Introduces changes:
- Enable building from macos `LOCAL_OS=macos make docker-build` by using `gsed` (`brew install gnu-sed`)
- Fix a `docker-for-mac` issue with shared directories (docker container mounts failed) not being canonicalize (because `ls -l /tmp -> private/tmp` in macos) by using `readlink` and `greadlink` (`brew install coreutils`)
- Download and verify Mozilla CA cert store in PEM format provided by `curl` [here](https://curl.haxx.se/docs/caextract.html) 
- Add the cert store to the docker image in an arbitrary chosen path from [this](https://golang.org/src/crypto/x509/root_linux.go) list

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,16 @@
 REGISTRY?=directxman12
 IMAGE?=k8s-prometheus-adapter
-TEMP_DIR:=$(shell mktemp -d)
 ARCH?=amd64
 ALL_ARCH=amd64 arm arm64 ppc64le s390x
 ML_PLATFORMS=linux/amd64,linux/arm,linux/arm64,linux/ppc64le,linux/s390x
 OUT_DIR?=./_output
 VENDOR_DOCKERIZED=0
+# You need to run "brew install coreutils gnu-sed" to use "macos" below
+LOCAL_OS?=linux
+SED_BIN?=sed
+READLINK_BIN?=readlink
+# Get Mozilla CA Cert store in PEM format from curl, https://curl.haxx.se/docs/caextract.html
+CA_CERTIFICATES_URL?=https://curl.haxx.se/ca/cacert.pem
 
 VERSION?=latest
 GOIMAGE=golang:1.10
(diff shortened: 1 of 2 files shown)


Vlaaaaaaad:
@DirectXMan12 any chance @nicolaevladescu can get a review on this? This is a blocker for us.

s-urbaniak:
@Vlaaaaaaad @nicolaevladescu I am a bit worried about adding official certificates to the image. These needs long-term maintenance and is usually not idiomatic in other kubernetes components.

As you recognized, there was `--prometheus-ca-file` added, but you can also use `--prometheus-auth-config` which lets you specify a kubernetes-style kubeconfig which can also include the whole  `*.remote.com` certificate chain.

So, I would rather recommend you to either a) synthesize a kubeconfig-style including the cert-chain, mount that into the container and specify `--prometheus-auth-config` or b) synthesize a PEM certificate chain, mount that into the container and specify `--prometheus-ca-file`.

As for the fixes to make this buildable from macos, do you mind to create a separate PR for this?

Vlaaaaaaad:
@DirectXMan12: hm... would modifying the helm chart at [stable/prometheus-adapter](https://github.com/helm/charts/tree/master/stable/prometheus-adapter) to mount the host's `/etc/ssl/certs/`( or a customizable path since this differs a bit) be a better idea? Looking into it that's what some charts seem to do.

nicolaevladescu:
@Vlaaaaaaad The host's `/etc/ssl/certs/` has the same Mozilla CA store but yes, i would simplify the long-term maintenance by doing it on the host side. Aside from that, maybe we can cherry pick just the `Makefile` macos improvements if you guys see any value in those for local development.

DirectXMan12:
My suggestion is in line with @s-urbaniak -- I'd recommend mounting in your cert bundle somehow.  Either with a Secret or ConfigMap is best -- you don't always know that you can or want to mount in the host's certs (different OS paths, different permissions, etc).  Injecting CA bundles into dockerfiles means you have to rebuild if your CA bundle changes, and that we'd have to rebuild on bundle changes.

I'll like @brancz and @s-urbaniak make the final decision though, since they're the primary maintainers of this repo at this point.

nicolaevladescu:
Thanks guys, i think we can manage without the CA part, i'll open another PR with the `Makefile` improvements only.

Issue #178, opened 2019-03-21
Node resource can not be accessed correctly


xjas:
**What happened:**

Using `kubectl get --raw "/apis/custom.metrics.k8s.io/v1beta1/" | jq . | less`, I get a unexpected response, as follows:
```
{
  "kind": "APIResourceList",
  "apiVersion": "v1",
  "groupVersion": "custom.metrics.k8s.io/v1beta1",
  "resources": [
    {
      "name": "nodes/http_requests",
      "singularName": "",
      "namespaced": true,
      "kind": "MetricValueList",
      "verbs": [
        "get"
      ]
    },
    {
      "name": "namespaces/http_requests",
      "singularName": "",
      "namespaced": false,
      "kind": "MetricValueList",
      "verbs": [
        "get"
      ]
    },
    {
      "name": "pods/http_requests",
      "singularName": "",
      "namespaced": true,
      "kind": "MetricValueList",
      "verbs": [
        "get"
      ]
    }
  ]
}
```

And also, using `kubectl get --raw "/apis/custom.metrics.k8s.io/v1beta1/node/*/http_requests" | jq . | less`, 
I get a unexpected response:
`Error from server (NotFound): the server could not find the metric http_requests for node`

**What you expected to happen:**

In my opinion, the resouce node should in the same hierarchy with resouce namespace, because in kubernetes they are non-namespaced apparently. And I have request the metric `http_requests` of the resource `namespace`, and it response as expected.

Personally, I think  the fileld `namespaced` should be false for resource `nodes/http_requests`, and the request `kubectl get --raw "/apis/custom.metrics.k8s.io/v1beta1/node/*/http_requests" should return the correct metric info.

**How to reproduce it (as minimally and precisely as possible):**
Just use the config rules I pasted and make requests: `kubectl get --raw "/apis/custom.metrics.k8s.io/v1beta1/" | jq . | less` and  `kubectl get --raw "/apis/custom.metrics.k8s.io/v1beta1/node/*/http_requests" | jq . | less`.

**Environment:**
- Kubernetes v1.9.6
- Adapter image: directxman12/k8s-prometheus-adapter-amd64:v0.4.0
- Adapter image: directxman12/k8s-prometheus-adapter-amd64:v0.4.1(build it with cmd `make build-local-image`)
- My config rule is as follows:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: adapter-config
  namespace: custom-metrics
data:
[...]

PR #179, opened 2019-03-21
Fix namespaced bug of node resource for issue #178

xjas:
Fix issue:[#178](https://github.com/DirectXMan12/k8s-prometheus-adapter/issues/178)

Fix namespaced bug of node resource, node resource should be a
non-namespaced resource as the resource namespace.

By doing this fix, node resouce can be accessed as expected, as the issue explains.

Closes issue:[#178](https://github.com/DirectXMan12/k8s-prometheus-adapter/issues/178)

### DIFF
--- a/pkg/custom-provider/metric_namer.go
+++ b/pkg/custom-provider/metric_namer.go
@@ -13,9 +13,12 @@ import (
 	"github.com/directxman12/k8s-prometheus-adapter/pkg/naming"
 )
 
-var nsGroupResource = schema.GroupResource{Resource: "namespaces"}
-var nodeGroupResource = schema.GroupResource{Resource: "nodes"}
-var groupNameSanitizer = strings.NewReplacer(".", "_", "-", "_")
+var (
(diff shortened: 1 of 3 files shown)


s-urbaniak:
Thanks a lot for the contribution! This PR and the referenced issue seem reasonable to me :+1: I restarted the build to see if it is flaky.

xjas:
> Thanks a lot for the contribution! This PR and the referenced issue seem reasonable to me  I restarted the build to see if it is flaky.

Thanks for your response.

Sorry for forgeting to do the test before making the PR, and I will check the reason why the ci failed.

xjas:
I think the CI problem have been solved in an ugly way.
However, the scheme I have coded is not graceful at all. If any namespaced resources are  introduced into k8s in the future, the code maybe have to be modified again.

I wonder, if there are any way to change the situation?

s-urbaniak:
@xjas I think the patch as-is is fine for now :+1: I can do a consolidated PR which unifies detection of namespaced objects in a follow-up. For now this LGTM.

PR #192, opened 2019-05-07
pkg/naming: fix LabelValuesByName rendering

### LINKED ISSUE #191: LabelValuesByName incorrect value
I am using LabelValuesByName in my query . The issue i am facing is when i try to get pod value from it , it is returning me an array instead of string with all the pods name 

Expected 
<<index .LabelValuesByName "pod_name">> 
"pod1|pod2"

Got :
[pod1 pod2]

s-urbaniak:
According to documentation LabelValuesByName is supposed to render "|" separated values.
Currently it is returned as a slice per label name. This fixes it

Fixes #191

@paulfantom: as discussed, let's add unit tests for the external metrics rendering too.

### DIFF
--- a/pkg/naming/metrics_query_test.go
+++ b/pkg/naming/metrics_query_test.go
@@ -1,0 +1,234 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package naming
+
+import (
+	"fmt"
+	"testing"
+
+	prom "github.com/directxman12/k8s-prometheus-adapter/pkg/client"
+	pmodel "github.com/prometheus/common/model"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type resourceConverterMock struct {
+	namespaced bool
+}
+
+// ResourcesForSeries is a mock that returns a single group resource,
+// namely the series as a resource itself.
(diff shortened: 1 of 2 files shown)


pkg/naming/metrics_query_test.go
```diff
@@ -232,3 +234,121 @@ func TestBuildSelector(t *testing.T) {
...
+		},
+		{
+			name: "multiple LabelMatchers value",
+
+			mq:             mustNewQuery(`<<.LabelMatchers>>`, false),
+			metricSelector: labels.NewSelector().Add(*singleReq, *singleReqMultVals),
+
+			check: checks(
+				hasError(nil),
+				hasSelector("foo=\"bar\",foo=~\"bar|baz\""),
```
s-urbaniak: nit use backticks for better readability:
```
`foo="bar",foo=~"bar|baz"`
```

pkg/naming/metrics_query_test.go
```diff
@@ -232,3 +234,121 @@ func TestBuildSelector(t *testing.T) {
...
+		},
+		{
+			name: "single LabelMatchers value",
+
+			mq:             mustNewQuery(`<<.LabelMatchers>>`, false),
+			metricSelector: labels.NewSelector().Add(*singleReq),
+
+			check: checks(
+				hasError(nil),
+				hasSelector("foo=\"bar\""),
```
s-urbaniak: nit: use backticks for better readabilty (see below)

pkg/naming/metrics_query_test.go
```diff
@@ -0,0 +1,354 @@
...
+			check: checks(
+				hasError(nil),
+				hasSelector("map[foo:bar]"),
+			),
+		},
+		{
+			name: "multiple LabelValuesByName",
+
+			mq:             mustNewQuery(`<<.LabelValuesByName>>`, false),
+			metricSelector: labels.NewSelector().Add(*singleReq, *singleReqMultVals),
```
s-urbaniak: as discussed, let's wrap this to have a bit cleaner code here :-)

pkg/naming/metrics_query_test.go
```diff
@@ -0,0 +1,354 @@
...
+
+			check: checks(
+				hasError(nil),
+				hasSelector("map[foo:bar|baz]"),
```
s-urbaniak: just to avoid confusion: you are adding two metric selectors here using the same key `"foo"`. Wouldn't it be better to have two separate keys so they appear separately in the map?

PR #214, opened 2019-06-23
Yaml typo for seriesFilters example

thisisamurray:
Small thing, but missing `-` in order to make seriesFilters an array. Ran into issues getting the example to run until I saw this https://github.com/DirectXMan12/k8s-prometheus-adapter/blob/master/docs/sample-config.yaml#L40

Just updating to reflect that structure


s-urbaniak:
good catch, thank you :+1: and sorry for not merging earlier as I have been on vacation.

PR #223, opened 2019-07-20
instrument k8s-prometheus-adapter to expose prometheus metrics

### LINKED ISSUE #218: Expose metrics about adapter operation
As far as I can tell, the adapter doesn't expose any metrics about itself? We've had multiple occasions where someone accidentally broke the adaper config with changes to other components (Prometheus, routing, ...), and it sometimes takes a while before someone notices since we don't have anything to alert on from the adapter.
So, I'd like to propose adding some metrics about the adapter operation. Some simple ones could be:
- Contact with Prometheus works (gauge 1/0)
- Timestamp of configuration
- Number of rules
- Number of requests for missing metrics

pdbogen:
fixes #218

there are no doubt more metrics that could be exposed here, but this seems like a good start.


pkg/config/loader.go
```diff
@@ -28,5 +30,7 @@ func FromYAML(contents []byte) (*MetricsDiscoveryConfig, error) {
 	if err := yaml.UnmarshalStrict(contents, &cfg); err != nil {
 		return nil, fmt.Errorf("unable to parse metrics discovery config: %v", err)
 	}
+	metrics.Rules.WithLabelValues("normal").Set(float64(len(cfg.Rules)))
```
s-urbaniak: nit: rather call this `"resource"`.

pkg/errors/errors.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/directxman12/k8s-prometheus-adapter/pkg/metrics"
+	"github.com/kubernetes-incubator/custom-metrics-apiserver/pkg/provider"
+	apierr "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func NewMetricNotFoundError(resource schema.GroupResource, metricName string) error {
+	metrics.Errors.WithLabelValues("not_found").Inc()
```
s-urbaniak: I see what this code is trying to achieve, namely to reduce reduce the repetition of metric updates in the caller code, but this construct is not idiomatic. This method should do what it implies, creating a new error and not introduce side effects.

Suggestion: let the `MetricNotFoundError` be a sentinal error and react on that in the calling code.

pkg/errors/errors.go
```diff
@@ -0,0 +1,18 @@
...
+	return provider.NewMetricNotFoundError(resource, metricName)
+}
+
+func NewInternalError(err error) *apierr.StatusError {
+	metrics.Errors.WithLabelValues("internal").Inc()
```
s-urbaniak: see comment above.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,38 @@
...
+}, []string{"type"})
+
+var Rules = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+	Namespace: MetricsNamespace,
+	Name:      "roles",
+	Help:      "number of configured rules",
+}, []string{"type"})
+
+func init() {
+	prometheus.MustRegister(
```
s-urbaniak: Let's not use `init()` and register the prometheus registry as part of package initialization. Let's rather intanciate the registry in `main.go`, and dependency inject it into the providers.

pkg/config/loader.go
pdbogen: happy to adjust. do you mean the specific label should be `resource` instead of `normal`?

pkg/errors/errors.go
pdbogen: The caller in most cases is somewhere inside `apiserver` package. I've got no qualms about not using utility functions for this, but it _is_ going to be a lot less DRY. I'll update the PR to reflect that, and you can decide if you prefer it to this option.

pkg/metrics/metrics.go
pdbogen: I feel _this_ is non-idiomatic for the Prometheus client SDK, but you're the boss. :)

It's going to be more complicated than just injecting a registry, because the registry doesn't actually provide access to the metrics registered to it. We'll have to inject either the metric of interest, or an object that wraps all of the metrics.

pkg/config/loader.go
s-urbaniak: Yes :+1: as this refers to "resource metrics".

pkg/metrics/metrics.go
s-urbaniak: Injecting the metric of interest sounds good to me :+1: 

Regarding dependency injection: Indeed, historically `init()` based registry initialization has been used a lot within the Go ecosystem but many Prometheus ecosystem maintainers recommend dependency injection (be it the metric or the registry itself) instead. Also see the metrics overhaul KEP [1] for the motivation of not using `init()` based metrics registration.

[1] https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/20181106-kubernetes-metrics-overhaul.md#motivation

pdbogen: yep, all good.

I'm happy to rework to inject just the metric, but I _did_ already push commits that refactor to inject a ServiceMetrics object. Are you OK with that approach, or do you feel strongly that it should only be prometheus' Gauge (etc.) object injected?

pkg/errors/errors.go
pdbogen: I've updated the PR to not introduce an `errors` package, preserving the previous behavior. I didn't see any opportunity to treat the error as a sentinel at a higher level, in a way that would improve clarify or reduce repetition.

(e.g., several of the functions are called in multiple places, so it would be more code & more complicated code farther from the actual causal location to check for certain types errors & report metrics about them.)

The places where we really _could_ reduce repetition are inside apiserver, not in this code base.

(Additionally, since the metrics are no longer global, each call site is even slightly more verbose to handle the case where, e.g. during testing, metrics have not been instantiated.)

Happy to adjust for any further feedback you might have.

pdbogen:
hi, @s-urbaniak. I think I addressed your comments back in July. I should've pinged earlier, but would you be willing to have another look? (If you have other requests, with fresh eyes, please don't hesitate to bring them up. Happy to continue working on this.)

machine424:
We're really interested in this.

junaid-ali:
Having some metrics exposed would also help understanding the resource usage under load.

PR #226, opened 2019-08-02
Fixed port name in service monitor

cgkades:
The way the service is created here creates a port named 80-8080 and not http. This causes this example to not work.


cgkades:
My commit is only in a MD file, and it looks like the build failed on connecting to github. Is there a way to retry?

dgrisonnet:
This should have been fixed by #379.

Issue #229, opened 2019-08-14
Series query 'start' timestamps lag 20 minutes behind current time


weibeld:
Using [v0.5.0](https://github.com/DirectXMan12/k8s-prometheus-adapter/releases/tag/v0.5.0), the timestamp of the `start` parameter of the `/api/v1/series` requests to Prometheus lags 20 minutes behind the current time.

From the logs:

```
I0814 12:54:05.324711  1 api.go:74] GET http://prometheus.custom-metrics.svc.cluster.local:9090/api/v1/series?match%5B%5D=autoscaling_http_requests_total&start=1565786045.323 200 OK
```

As you can see, the current time is 12:54:05 GMT, but the `start` timestamp is 1565786045.323, which is 12:34:05 GMT.

The relist interval is set to 10 seconds. As far as I know, this means that the `start` timestamp should be 10 seconds behind the current time (get time series that have been updated since the previous series query). However, it is always 20 minutes behind the current time, no matter what the relist interval is.

The effect of this is that when metrics disappear from Prometheus (e.g. Pods are deleted), they are still reported in`kubectl get --raw '/apis/custom.metrics.k8s.io/v1beta1'` until 20 minutes after the metric disappeared.

For the `/api/v1/query` request (i.e. when a user makes a request to the Custom Metrics API), the `time` parameter timestamps are correct (i.e. correspond to the current time).

weibeld:
In [custom-provider/provider.go:220](https://github.com/DirectXMan12/k8s-prometheus-adapter/blob/master/pkg/custom-provider/provider.go#L220) the start time is set to `currentTime - maxAge`, and `maxAge` is passed in [here](https://github.com/DirectXMan12/k8s-prometheus-adapter/blob/master/cmd/adapter/adapter.go#L170) from a command-line flag named `MetricsMaxAge`, which has a [default value of 20 minutes](https://github.com/DirectXMan12/k8s-prometheus-adapter/blob/master/cmd/adapter/adapter.go#L245).

So, on [custom-provider/provider.go:220](https://github.com/DirectXMan12/k8s-prometheus-adapter/blob/master/pkg/custom-provider/provider.go#L220), shouldn't it be `currentTime - updateInterval` instead?

weibeld:
I see that #136 introduced the `metrics-max-age` flag and changed

```
startTime := pmodel.Now().Add(-1 * l.updateInterval)
```

to

```
[...]

PR #230, opened 2019-08-14
Set --metrics-max-age default value equal to relist interval

weibeld:
The`--metrics-max-age` flag has been introduced in #136 with a fixed default value of 20 minutes. 

The current default value leads to the situation that when metrics drop out of Prometheus, they are still reported in the discovery information of the Custom Metrics API (`/apis/custom.metrics.k8s.io/v1beta1`) for 20 minutes.

This can be quite confusing if you don't know about the `--metrics-max-age` flag and expect the behaviour that is described in the docs. Namely that the Prometheus Adapter discovers all metrics that have been updated since the last discovery query (defined by the relist interval), which is also the reason the relist interval must be at least as large as the Prometheus scraping interval (see #229).

By setting the default value of `--metrics-max-age` equal to the relist interval, the expected behaviour as described in the docs is preserved, but you can still set `--metrics-max-age` to some other value if you know about it.

The other way to go would be to update the docs and expect all users to set the `--metrics-max-age` flag (and the [Helm chart](https://github.com/helm/charts/blob/master/stable/prometheus-adapter/templates/custom-metrics-apiserver-deployment.yaml) would need to be updated to use it too). But this would make the configuration of the Prometheus Adapter even more complicated than it already is. It is probably easier to just remember to set the relist interval at least as large as the Prometheus scraping interval, than to additionally have to figure out what the `metrics-max-age` is and what to set it to.

### DIFF
--- a/cmd/adapter/adapter.go
+++ b/cmd/adapter/adapter.go
@@ -250,8 +250,8 @@ func main() {
 		klog.Fatalf("unable to parse flags: %v", err)
 	}
 
-	// if --metrics-max-age is not set, make it equal to --metrics-relist-interval 
-	if cmd.MetricsMaxAge == 0 * time.Second {
+	// if --metrics-max-age is not set, make it equal to --metrics-relist-interval
+	if cmd.MetricsMaxAge == 0*time.Second {
 		cmd.MetricsMaxAge = cmd.MetricsRelistInterval
 	}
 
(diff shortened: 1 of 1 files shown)


s-urbaniak:
Despite the age of the PR, I think this is still reasonable, hence merging.

PR #252, opened 2019-11-04
Support Cross Namespace External Metrics

AshMenhennett:
**Context**: 
When retrieving metrics from namespace B and intending to scale with HPA Object in namespace A, the metrics will have a namespace label selector appended with Namespace A. This breaks the ability to scale based on metrics in a different namespace.

This PR adds support for this by adding an `options` field for a `DiscoveryRule`, allowing an operator to opt-in to this functionality, defaulting to the existing behaviour. 

This PR also adds logic to avoid any 64-bit Integer overflow issues when metrics retrieved from Prometheus are of the form `+Inf`. When there is an overflow, the value would return `math.MinInt64`, which is negative. Observations are that when scaling needs to happen (Prometheus returns with an `+Inf` metric value) scaling would not occur as the metrics value was a negative Integer. 

To provide a more deterministic outcome for when a metric value of `-Inf` is retrieved from Prometheus, the value `math.MinInt64` will be used, rather than relying on an overflow.

**Validations**: 
Unit tests have been added to assert that setting the `DetatchFromNamespace` in the options object results in the correct functionality. This has also been through reasonable acceptance testing by myself.
I have not been able to replicate a scenario where Prometheus returns with an `-Inf` metric value, but have been able to replicate the `+Inf` metric value scenario successfully.

**Links**: 
+ The following link is a point of reference to the issue that this PR solves: https://github.com/DirectXMan12/k8s-prometheus-adapter/issues/195

### DIFF
--- a/pkg/naming/metrics_query_test.go
+++ b/pkg/naming/metrics_query_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 
 	prom "github.com/directxman12/k8s-prometheus-adapter/pkg/client"
+	"github.com/directxman12/k8s-prometheus-adapter/pkg/config"
 	pmodel "github.com/prometheus/common/model"
 	labels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -261,6 +262,7 @@ func TestBuildExternalSelector(t *testing.T) {
 		groupBy        string
 		groupBySlice   []string
 		metricSelector labels.Selector
+		options        config.Options
 
 		check checkFunc
 	}{
@@ -341,6 +343,36 @@ func TestBuildExternalSelector(t *testing.T) {
 			),
 		},
 		{
+			name: "single LabelMatchers value with namespace - detached namespace",
+
+			mq:        mustNewQuery(`<<.LabelMatchers>>`),
+			namespace: "default",
+			options:   config.Options{DetatchFromNamespace: true},
+			metricSelector: labels.NewSelector().Add(
+				*mustNewLabelRequirement("foo", selection.Equals, []string{"bar"}),
+			),
+
+			check: checks(
+				hasError(nil),
+				hasSelector(`foo="bar"`),
+			),
+		},
+		{
+			name: "single LabelMatchers value with namespace - explicit attached namespace",
+
+			mq:        mustNewQuery(`<<.LabelMatchers>>`),
+			namespace: "default",
+			options:   config.Options{DetatchFromNamespace: false},
+			metricSelector: labels.NewSelector().Add(
+				*mustNewLabelRequirement("foo", selection.Equals, []string{"bar"}),
+			),
+
+			check: checks(
+				hasError(nil),
+				hasSelector(`foo="bar",namespaces="default"`),
+			),
+		},
+		{
 			name: "multiple LabelMatchers value",
 
 			mq: mustNewQuery(`<<.LabelMatchers>>`),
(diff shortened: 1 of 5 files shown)


AshMenhennett:
Hi @DirectXMan12 ,

How are we looking to get this guy into master in the next couple weeks?
This feature opens up a world of possibilities from my perspective.

Regards,
Ashley

s-urbaniak:
@AshMenhennett thanks a lot for the contribution. @DirectXMan12 is not working on this project any more and I took over maintenance. Please give me some time for a review.

AshMenhennett:
Sure! Thanks for your work :)

AshMenhennett:
Hey @s-urbaniak,

Any update on this one?
Would be a good one to get in.

Regards,
Ashley

s-urbaniak:
@AshMenhennett sorry for the long delay. At first eye-sight I am fine with this change, but want to make sure that from the perspective of k8s we are doing "the right thing".

pinging @brancz for advise.

pkg/naming/metrics_query_test.go
```diff
@@ -340,6 +342,36 @@ func TestBuildExternalSelector(t *testing.T) {
...
+			),
+		},
+		{
+			name: "single LabelMatchers value with namespace - explicit attached namespace",
+
+			mq:        mustNewQuery(`<<.LabelMatchers>>`),
+			namespace: "default",
+			options:   config.Options{DetatchFromNamespace: false},
+			metricSelector: labels.NewSelector().Add(
+				*mustNewLabelRequirement("foo", selection.Equals, []string{"bar"}),
```
brancz: can you add a test where a different namespace label is provided?

AshMenhennett: Good call. Have added additional test case to ensure there is no 'default' namespace applied when `config.Options{DetatchFromNamespace} == false`.

brancz: Can you add another one where the user specifies a label requirement for the namespace label, and we ensure that it's truly enforced?

aleks-fofanov:
Hi @s-urbaniak , @brancz 
Can we please have your review here?

brancz:
It might have been hidden, but requested another change here: https://github.com/DirectXMan12/k8s-prometheus-adapter/pull/252#discussion_r370548735

pkg/naming/metrics_query_test.go
AshMenhennett: Hi @brancz ,
Apologies for the delay..

Had a think about your request, my understanding is that you would like to ensure that when a user specifies a Label requirement that has key `namespace`, that this should also be dropped while forming the selector.

In this case, I would suggest not enforcing this.

**Reasoning**:
The original intention of this feature is to allow cross namespace metrics, ensuring that metrics are not restricted to the namespace to which the HPA resource  resides.
If we had a scenario where we are doing cross-namespace metrics, and the metric we are retrieving happens to have a `namepace` label that we need to target (for whatever reason) then, enforcing detached namespace in this scenario would prevent the user from fulfilling their requirements.

What are your thoughts?

s-urbaniak: @AshMenhennett I am very sorry for the long delay, but I am still having doubts here. While I agree that external metrics are not bound to namespaces, having an explicit option to allow inter-namespace cross-correlation still sounds like a leaky abstraction.

Wouldn't be your problem be solvable by simply adding a Prometheus recording rule to your stack which cross-correlates namespaces and reference that in the adapter configuration?

brancz: The external metrics API was intentionally specified to be bound to a namespace, in this case it doesn't map 100% perfectly against Prometheus norms, but I think violating this would be a mistake. If we can handle this with a recording rule instead I do agree that I would prefer that approach.

AshMenhennett: Thanks for the timely reply guys!

I definitely see your concerns and understand the intention of the specification and the need to stay inline with it.

As a tactical solution, Prometheus recording rules could be used, however, as a more strategic solution, the ability to allow fine-grained control on a per resource case would be the primary goal/ benefit.

For example, in a large enough setting, this would require recording rules and the queries to be maintained for each scaling resource, each metric source and each environment. This isn't a massive issue, but I'm sure you can see what I'm getting at.

Thoughts?

AshMenhennett: Are we at consensus of leveraging recording rules rather than introducing this feature?

s-urbaniak: @AshMenhennett: yes, that is our current suggested solution. Please let me know if this works out for you. I am really in favor of a recording rule based solution as we would not have to introduce additional code paths.

Another data point is that configuration of prometheus-adapter is already complex enough, hence I am voting to add code if and only if we have no workaround available in Prometheus configuration.

PR #258, opened 2019-11-13
Small typo in config example snippet

rjackson90:
This was a small gotcha for me as I was setting up my configuration.


s-urbaniak:
good catch, thanks!

PR #286, opened 2020-04-13
Use build args to fix building on macos

johanneswuerbach:
A different approach to resolved build issues on macos

Fixes https://github.com/DirectXMan12/k8s-prometheus-adapter/issues/198
Fixes https://github.com/DirectXMan12/k8s-prometheus-adapter/pull/231
Fixes https://github.com/DirectXMan12/k8s-prometheus-adapter/pull/244


johanneswuerbach:
Do you have an example of what you had in mind? The make targets shouldn’t require any manual setting and work as before. Are there existing docs, which need updating?

lilic:
The directory just needs to be created if it doesn't exist, at least for me locally.

johanneswuerbach:
@lilic thanks, the output directory is created now if it doesn't exists.

lilic:
Thanks! lgtm @s-urbaniak PTAL

s-urbaniak:
lgtm :+1:

Issue #274, opened 2020-03-02
There are 403 logs after deployment


guleng:
```
I0302 07:57:40.557744       1 round_trippers.go:438] GET https://10.68.0.1:443/api/v1/pods?resourceVersion=1483856&timeout=8m9s&timeoutSeconds=489&watch=true 403 Forbidden in 2 milliseconds
I0302 07:57:40.557792       1 round_trippers.go:444] Response Headers:
I0302 07:57:40.557801       1 round_trippers.go:447]     Content-Type: application/json
I0302 07:57:40.557806       1 round_trippers.go:447]     X-Content-Type-Options: nosniff
I0302 07:57:40.557812       1 round_trippers.go:447]     Content-Length: 299
I0302 07:57:40.557817       1 round_trippers.go:447]     Date: Mon, 02 Mar 2020 07:57:40 GMT
I0302 07:57:40.557970       1 request.go:947] Response Body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"pods is forbidden: User \"system:serviceaccount:custom-metrics:custom-metrics-apiserver\" cannot watch resource \"pods\" in API group \"\" at the cluster scope","reason":"Forbidden","details":{"kind":"pods"},"code":403}
E0302 07:57:40.558086       1 reflector.go:283] k8s.io/client-go/informers/factory.go:133: Failed to watch *v1.Pod: unknown (get pods)
I0302 07:57:40.630759       1 reflector.go:160] Listing and watching *v1.Node from k8s.io/client-go/informers/factory.go:133
[...]

PR #290, opened 2020-05-14
fix forbidden issue #274 by adding appropriate verbs in cluster role

nirrozenbaum:
link to the issue: 
https://github.com/DirectXMan12/k8s-prometheus-adapter/issues/274#issuecomment-602282471

### DIFF
--- a/deploy/manifests/custom-metrics-resource-reader-cluster-role.yaml
+++ b/deploy/manifests/custom-metrics-resource-reader-cluster-role.yaml
@@ -6,9 +6,10 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - namespaces
   - pods
-  - services
+  - nodes
+  - nodes/stats
   verbs:
   - get
   - list
+  - watch


deploy/manifests/custom-metrics-resource-reader-cluster-role.yaml
```diff
@@ -6,9 +6,10 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - namespaces
   - pods
-  - services
```
s-urbaniak, line 11: why do we need to remove namespaces and services?

nirrozenbaum: this setup is the one that is used in the helm charts. please see here:
https://github.com/helm/charts/blob/master/stable/prometheus-adapter/templates/resource-metrics-cluster-role.yaml

in addition I've had this bug #274, and after setting the cluster role as suggested in this pull request everything worked perfectly.

no need for namespaces and services.

s-urbaniak:
LGTM, thank you!

PR #304, opened 2020-07-22
add endTime variable for promeClient.Series api

DanielQujun:
Currently, when Prometheus adapter querying Prometheus series API to updating metrics, It's hardcoded set `endtime` parameter to 0, which means we won't set endtime parameter in prometheus query URL

https://github.com/DirectXMan12/k8s-prometheus-adapter/blob/master/pkg/client/api.go#L152
```
	if interval.End != 0 {
		vals.Set("end", interval.End.String())
	}
```

then Prometheus will set A very large default time as `endtime` to query its backend storage, this might be failed, and return no metrics data.

I add value to `endtime`, It might alleviate this potential issue. 

related issue: https://github.com/m3db/m3/issues/2471

Signed-off-by: 屈骏 <qujun@tiduyun.com>

### DIFF
--- a/pkg/custom-provider/provider.go
+++ b/pkg/custom-provider/provider.go
@@ -218,6 +218,7 @@ type selectorSeries struct {
 
 func (l *cachingMetricsLister) updateMetrics() error {
 	startTime := pmodel.Now().Add(-1 * l.maxAge)
+	endTime := pmodel.Now().Add(1 * l.maxAge)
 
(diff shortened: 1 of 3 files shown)


DanielQujun:
@s-urbaniak Hi, PTAL

s-urbaniak:
Sorry for the lag, mostly due to summer vacation :blush: Looking at this now.

PR #308, opened 2020-08-11
Fix NaN not to be cast to int64

Ladicle:
ref: https://github.com/DirectXMan12/k8s-prometheus-adapter/issues/280

This pull-request prevents the NaN from being cast to int64. This is because the value of histogram_quantile(), which returns a NaN when the request is 0, such as calculating a custom metric for 95%tile, will overflow and returned `-9223372036854775808m`.


s-urbaniak:
LGTM, this is reasonable, thank you!

PR #330, opened 2020-10-29
Populate both CPU and Memory resource container metrics if one is specified

joelsmith:
We're experiencing an issue where HPA won't work with pods that have init containers. The root cause seems to be that the init containers have memory metrics but no CPU metrics, and HPA refuses to take action if any containers in the pod have container metrics, but are missing either CPU or memory.

This patch makes the adapter metrics look like the metrics server endpoint.

Before:
```
items:
- apiVersion: metrics.k8s.io/v1beta1
  containers:
  - name: gerbil
    usage:
      cpu: "0"
      memory: 1568Ki
  - name: gerbil-init
    usage:
      memory: "0"
  kind: PodMetrics
```
After:
```
items:
- apiVersion: metrics.k8s.io/v1beta1
  containers:
  - name: gerbil
    usage:
      cpu: "0"
      memory: 1568Ki
  - name: gerbil-init
    usage:
      cpu: "0"
      memory: "0"
  kind: PodMetrics
```

### DIFF
--- a/pkg/resourceprovider/provider.go
+++ b/pkg/resourceprovider/provider.go
@@ -226,6 +226,17 @@ func (p *resourceProvider) assignForPod(pod apitypes.NamespacedName, resultsByNs
 		}
 	}
 
+	// check for any containers that have either memory usage or CPU usage, but not both
+	for _, containerMetric := range containerMetrics {
+		_, hasMemory := containerMetric.Usage[corev1.ResourceMemory]
+		_, hasCPU := containerMetric.Usage[corev1.ResourceCPU]
+		if hasMemory && !hasCPU {
+			containerMetric.Usage[corev1.ResourceCPU] = *resource.NewMilliQuantity(int64(0), resource.BinarySI)
+		} else if hasCPU && !hasMemory {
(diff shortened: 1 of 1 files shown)


s-urbaniak:
lgtm, thank you for the contribution!

s-urbaniak:
@joelsmith general question though: we should recheck if metrics-server behaves the same and clarify if the behavior is unique across all resource metrics server implementations.

joelsmith:
@s-urbaniak while the implementation is different, metrics server does indeed provide both CPU and memory for the init container in this case. I verified on an OCP cluster by disabling the prometheus adapter and replacing it with metrics server.  While metrics server does not have special code in it to populate missing CPU or memory values, the kubelet stats summary API does. Since the kubelet metrics add the missing metrics, then metrics server has them. Since the prometheus adapter doesn't use the stats summary API (instead getting metrics from prometheus with scrapes the values from the kubelet's cadvisor endpoint), it did not have the zero-valued CPU or memory metrics for init containers. Thanks for merging!

In case you are interested, here's the PR that we added some time ago to fix the issue in the metrics path that metrics server uses:
https://github.com/kubernetes/kubernetes/pull/88734/files#diff-89caf51d2d7b9fcbc06bf5c7c3aca05867afbbcfb8e1b4482d8237c7169335c5R45

PR #336, opened 2020-11-15
Add kustomization.yaml

sim1:
This allows to use the repo as a remote base with kustomize

### DIFF
--- a/kustomization.yml
+++ b/kustomization.yml
@@ -1,0 +1,14 @@
+kind: Kustomization
+apiVersion: kustomize.config.k8s.io/v1beta1
+resources:
+  - deploy/manifests/custom-metrics-apiserver-auth-delegator-cluster-role-binding.yaml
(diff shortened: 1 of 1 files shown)


kustomization.yml
```diff
@@ -0,0 +1,14 @@
+kind: Kustomization
```
MrD2, line 1: You should make this optional. Installation will fail if kustomize.config.k8s.io/v1beta1 does not exist

dgrisonnet:
Hello @sim1, thank you for your contribution! :smiley:

Just a small nit from my side, would you mind renaming the file to kustomization.y**a**ml, just so that we are consistent with the other manifests?

/cc @s-urbaniak

s-urbaniak:
@sim1 can you verify you signed the CNCF CLA? we are in the process of moving this project into kubernetes-sigs so I want to make sure we have the necessary prerequisites.

PR #344, opened 2020-12-02
Make NodeMetrics and PodMetrics APIs match K8s conventions

dgrisonnet:
Include the following fix to the NodeMetrics and PodMetrics APIs to match k8s
conventions: https://github.com/kubernetes-sigs/metrics-server/pull/654
- NodeMetrics and PodMetrics now have labels as they can be filtered by
them
- Field selector is now applied to NodeMetrics and PodMetrics instead
of Nodes and Pods

cc @s-urbaniak

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
 	k8s.io/kube-openapi v0.0.0-20200805222855-6aeccd4b50c6
 	k8s.io/metrics v0.19.3
 	k8s.io/sample-apiserver v0.18.5
-	sigs.k8s.io/metrics-server v0.3.7-0.20201028092756-2a1d1385123b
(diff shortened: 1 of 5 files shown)


s-urbaniak:
very nice, so we didn't have to do any code adjustments :tada:

s-urbaniak:
@dgrisonnet question: did you have a chance to verify custom-metrics-apiserver?

dgrisonnet:
Not yet, I was planning to look into it after fixing the metrics API here. I'll look into it now

dgrisonnet:
@s-urbaniak quick update, I was able to reproduce the same bug with the custom-metrics and external-metrics APIs? I'll prepare a PR to fix these in custom-metrics-apiserver.

PR #348, opened 2020-12-10
Populate metric selector for custom metrics

dgrisonnet:
When querying custom-metrics, the metric label selectors weren't populated in the resulting values.

A query with a metricLabelSelector query parameter set to `endpoint=https` was resulting in the following MetricValueList:
```
{
  "kind": "MetricValueList",
  "apiVersion": "custom.metrics.k8s.io/v1beta1",
  ...
  "items": [
    {
      ...
      "selector": null
    }
    ...
}
```

With this change, it now appears as:
```
{
  "kind": "MetricValueList",
  "apiVersion": "custom.metrics.k8s.io/v1beta1",
  ...
  "items": [
    {
      ...
      "selector": {
        "matchLabels": {
          "endpoint": "https"
        }
      }
    }
    ...
}
```

/cc @s-urbaniak

### DIFF
--- a/pkg/custom-provider/provider.go
+++ b/pkg/custom-provider/provider.go
@@ -78,7 +78,7 @@ func NewPrometheusProvider(mapper apimeta.RESTMapper, kubeClient dynamic.Interfa
 	}, lister
 }
 
-func (p *prometheusProvider) metricFor(value pmodel.SampleValue, name types.NamespacedName, info provider.CustomMetricInfo) (*custom_metrics.MetricValue, error) {
+func (p *prometheusProvider) metricFor(value pmodel.SampleValue, name types.NamespacedName, info provider.CustomMetricInfo, metricSelector labels.Selector) (*custom_metrics.MetricValue, error) {
 	ref, err := helpers.ReferenceFor(p.mapper, name, info)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


pkg/custom-provider/provider.go
```diff
@@ -90,18 +90,29 @@ func (p *prometheusProvider) metricFor(value pmodel.SampleValue, name types.Name
...
+			return nil, err
+		}
+		metric.Metric.Selector = sel
+	}
+
+	return metric, nil
 }
 
-func (p *prometheusProvider) metricsFor(valueSet pmodel.Vector, info provider.CustomMetricInfo, namespace string, names []string) (*custom_metrics.MetricValueList, error) {
+func (p *prometheusProvider) metricsFor(valueSet pmodel.Vector, namespace string, names []string, info provider.CustomMetricInfo, metricSelector labels.Selector) (*custom_metrics.MetricValueList, error) {
```
dgrisonnet, line 115: note: I changed the order of the parameters here to match the `metricFor` function

s-urbaniak:
lgtm, thank you!

s-urbaniak:
@dgrisonnet just because i simply forgot: as a follow-up we could potentially add some tests, wdyt?

dgrisonnet:
yes, that would definitely be a great addition. However, since I am not too familiar with prometheus-adapter test framework, it might take me some times.

Issue #324, opened 2020-10-05
Feature Proposal: Allow External Metrics to return metrics for a different namespace


carsonoid:
## Issue

I have a single [nsq](https://nsq.io/) installation in a K8s namespace named `nsq`.  But I have many deployments in different namespaces that need to autoscale using a topic and/or channel depth from the central installation.

The [external metrics proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/instrumentation/external-metrics-api.md#api) requires that a namespace is in the path. And the HPA will always put it's own namespace into the external metrics query.

This makes it impossible to use the adapter to scale a pod based on metrics from a different namespace than the HPA.

## Proposed Solution

Allow a special target label (`metrics_target_namespace`) to override the namespace that will be used in the Prometheus query.

Example HPA:

```yaml
apiVersion: autoscaling/v2beta2
kind: HorizontalPodAutoscaler
metadata:
  name: example
  namespace: example
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: example
  minReplicas: 1
  maxReplicas: 10
  metrics:
  - type: External
    external:
      metric:
        name: nsq_topic_depth
        selector:
          labelSelector:
            topic: my-topic
            metric_target_namespace: nsq
      target:
        type: Value
        AverageValue: 50
```

So with the example above, even though the HPA is in the `example` namespace. It will scale based on the topic in the `nsq` namespace. This simple override allows much greater flexibility for metrics while still fitting withing the Kubernetes API spec.

I have a fork of the adapter with this code running in production and would be happy to upstream the changes if the feature proposal is accepted.  I have only implemented the change in `external` metrics but it could also be implemented for all the other metric providers supported by the adapter.

h4ckroot:
I have a very similar use case, and I wonder if there is any workaround for this !.

carsonoid:
@h4ckroot  There is! I have already written code to do this. But I want to make sure that the proposed design is acceptable to @DirectXMan12

jackwhelpton:
I'd also be interested in something exactly like this, and am curious if there's a workaround until your code lands. I've tried several things over the last day or so: I wondered if I could use a cluster-scoped CRD as the target of a custom metric, but so far nothing seems to have worked out.

s-urbaniak:
[...]

PR #380, opened 2021-03-05
Allow metrics to be defined as `namespaced: false`

carsonoid:
When set to false, no namespace label will be set by the adapter based on the namespace
portion of the url in the request path.

This allows individual consumers to set namespace independent of the source kubernetes resource.

---

Example:

Given an adapter config like this:

```
    externalRules:
    - seriesQuery: 'nsq_topic_depth'
      resources:
        namespaced: false
```

An HPA could target a different namespace by setting it in the selector:

```
  - type: External
    external:
      metric:
        name: nsq_topic_depth
        selector:
          labelSelector:
            topic: my-topic
            namespace: nsq
```

This is useful for scaling on metrics from services that run in a differnt namespace than the source resource.

### DIFF
--- a/docs/sample-config.yaml
+++ b/docs/sample-config.yaml
@@ -65,5 +65,18 @@ rules:
       brand: {group: "cheese.io", resource: "brand"}
   metricQuery: 'count(cheddar{sharp="true"})'
 
+# external rules are not tied to a Kubernetes resource and can reference any metric
+# https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#autoscaling-on-metrics-not-related-to-kubernetes-objects
+external:
+- seriesQuery: '{__name__="queue_consumer_lag",name!=""}'
+  metricsQuery: sum(<<.Series>>{<<.LabelMatchers>>}) by (name)
+- seriesQuery: '{__name__="queue_depth",topic!=""}'
+  metricsQuery: sum(<<.Series>>{<<.LabelMatchers>>}) by (name)
+  # Kubernetes metric queries include a namespace in the query by default
+  # but you can explicitly disable namespaces if needed with "namespaced: false"
+  # this is useful if you have an HPA with an external metric in namespace A
+  # but want to query for metrics from namespace B
+  namespaced: false
+
 # TODO: should we be able to map to a constant instance of a resource
 # (e.g. `resources: {constant: [{resource: "namespace", name: "kube-system"}}]`)?
--- a/docs/sample-config.yaml
+++ b/docs/sample-config.yaml
@@ -76,7 +76,8 @@ external:
   # but you can explicitly disable namespaces if needed with "namespaced: false"
   # this is useful if you have an HPA with an external metric in namespace A
   # but want to query for metrics from namespace B
-  namespaced: false
+  resources:
+    namespaced: false
 
 # TODO: should we be able to map to a constant instance of a resource
 # (e.g. `resources: {constant: [{resource: "namespace", name: "kube-system"}}]`)?
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -58,6 +58,8 @@ type ResourceMapping struct {
 	// Overrides specifies exceptions to the above template, mapping label names
 	// to group-resources
 	Overrides map[string]GroupResource `json:"overrides,omitempty" yaml:"overrides,omitempty"`
+	//Namespaced ignores the source namespace of the requester and requires one in the query
+	Namespaced *bool `json:"namespaced,omitempty" yaml:"namespaced,omitempty"`
 }
 
 // GroupResource represents a Kubernetes group-resource.
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -58,7 +58,7 @@ type ResourceMapping struct {
 	// Overrides specifies exceptions to the above template, mapping label names
 	// to group-resources
 	Overrides map[string]GroupResource `json:"overrides,omitempty" yaml:"overrides,omitempty"`
-	//Namespaced ignores the source namespace of the requester and requires one in the query
+	// Namespaced ignores the source namespace of the requester and requires one in the query
 	Namespaced *bool `json:"namespaced,omitempty" yaml:"namespaced,omitempty"`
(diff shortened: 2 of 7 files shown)


carsonoid:
I signed it

carsonoid:
Fixes #324

carsonoid:
I could also put the `namespaced` config detection into the `metricsQuery` struct if that is preferred.

pkg/config/config.go
```diff
@@ -58,6 +58,8 @@ type ResourceMapping struct {
 	// Overrides specifies exceptions to the above template, mapping label names
 	// to group-resources
 	Overrides map[string]GroupResource `json:"overrides,omitempty" yaml:"overrides,omitempty"`
+	//Namespaced ignores the source namespace of the requester and requires one in the query
```
dgrisonnet, line 61: nit
```suggestion
	// Namespaced ignores the source namespace of the requester and requires one in the query
```

pkg/naming/metric_namer.go
```diff
@@ -102,10 +102,15 @@ type metricNamer struct {
 	nameMatches    *regexp.Regexp
 	nameAs         string
 	seriesMatchers []*ReMatcher
+	namespaced     *bool
 
 	ResourceConverter
 }
 
+func (n *metricNamer) isNamespaced() bool {
+	return (n.namespaced != nil) && *n.namespaced
```
dgrisonnet, line 111: IMO, we should still enforce the `namespace` label by default and only drop the enforcement if `namespaced: false`.

dgrisonnet (commented):
Thank you for looking into this @carsonoid! :slightly_smiling_face: 

Would you mind adding a test and some documentation?

dgrisonnet:
> I could also put the namespaced config detection into the metricsQuery struct if that is preferred.

Yes that would be preferable as it would allow to move the check in `BuildExternal`.

pkg/naming/metrics_query.go
```diff
@@ -88,7 +90,7 @@ type queryPart struct {
 func (q *metricsQuery) Build(series string, resource schema.GroupResource, namespace string, extraGroupBy []string, metricSelector labels.Selector, names ...string) (prom.Selector, error) {
 	queryParts := q.createQueryPartsFromSelector(metricSelector)
 
-	if namespace != "" {
+	if q.namespaced && namespace != "" {
```
dgrisonnet, line 93: we don't want to add the check here as it would enable this feature for the other metrics API

pkg/naming/metrics_query_test.go
```diff
@@ -76,7 +76,15 @@ func checks(cs ...checkFunc) checkFunc {
 
 func TestBuildSelector(t *testing.T) {
 	mustNewQuery := func(queryTemplate string, namespaced bool) MetricsQuery {
-		mq, err := NewMetricsQuery(queryTemplate, &resourceConverterMock{namespaced})
+		mq, err := NewMetricsQuery(queryTemplate, &resourceConverterMock{namespaced}, true)
```
dgrisonnet, line 79: same as above, we only want to check the `namespaced` value for BuildExtrernal as this feature is only meant for external metrics.

docs/sample-config.yaml
```diff
@@ -65,5 +65,18 @@ rules:
...
+external:
+- seriesQuery: '{__name__="queue_consumer_lag",name!=""}'
+  metricsQuery: sum(<<.Series>>{<<.LabelMatchers>>}) by (name)
+- seriesQuery: '{__name__="queue_depth",topic!=""}'
+  metricsQuery: sum(<<.Series>>{<<.LabelMatchers>>}) by (name)
+  # Kubernetes metric queries include a namespace in the query by default
+  # but you can explicitly disable namespaces if needed with "namespaced: false"
+  # this is useful if you have an HPA with an external metric in namespace A
+  # but want to query for metrics from namespace B
+  namespaced: false
```
dgrisonnet, line 79: this should be in `resources`

dgrisonnet (commented):
A couple of comments, otherwise it looks good to me. You did a great job with the documentation! :100:

pkg/naming/metrics_query_test.go
carsonoid: Agreed. But both both forms of metrics use the same type which is why all of the calls the `NewMetricsQuery` have to take a namespace.

We could do something else though:
1. Split `MetricsQuery` into two types: `MetricsQuery` & `ExternalMetricsQuery` . Each with it's own `NewX` function. This would be the cleanest but will result in more code changes
2. Add setter func for `namespaced` so we can change it on after calling `NewMetricsQuery`. This would work but doesn't feel like idomatic go
3. Leave the bare bool in `NewMetricsQuery` as-is

I'm fine with whatever you choose.

carsonoid:
@dgrisonnet  thanks for the review! I've made most of the changes requested. Let me know what you want to do about the last one.

pkg/naming/metrics_query_test.go
dgrisonnet: Splitting into two separate constructors sounds like the best option here.

carsonoid:
@dgrisonnet  I split the constructors as discussed :)

dgrisonnet:
Awesome @carsonoid! Thank you for keeping up with my requests, the code looks great :)

/lgtm
/approve

SkullTech:
@dgrisonnet Thanks for the merge!! I need this feature as soon as possible, I'd be very grateful if you can make a new release out of this now.

dgrisonnet:
There are a couple other changes that I'd like to have in the next minor release of prometheus-adapter so it might take a bit of time before the actual cut. In the meantime, you can still build an image with the latest changes by running `make docker-build` on the master branch.

SkullTech:
I see. Is there any way I can use the latest image built by me off the master branch with helm? If not, how can I go about deploying it, is there any documentation on that? The README just mentions the helm way. Thanks btw.

dgrisonnet:
I am not familiar with Helm, I generally use https://github.com/prometheus-operator/kube-prometheus. In the customize section of kube-prometheus README, you should be able to see how to customize the default configuration to change the image to the one you built yourself.

PR #405, opened 2021-06-01
Filter non-running pods

### LINKED ISSUE #402: Prometheus-adapter is collecting metrics for completed pods
Running prometheus-adapter with the resource metrics API configured results in collecting metrics for pods in a completed/succeeded state.

This is an issue that comes from metrics-server, one of prometheus-adapter dependency and it has been recently fixed recently in https://github.com/kubernetes-sigs/metrics-server/pull/651. This change should be part of the next release of metrics-server, v0.5.0, but the PodMetadataLister package of metrics-server which contains the fix isn't exported. So we will need to update our informer to only watch for pods in `running` state. Long term, it might be better to also only watch for pods metadata, but for now, we can stick to watching pods.

Note that it might be best to wait for https://github.com/kubernetes-sigs/prometheus-adapter/pull/399 to be merged as it bumps metrics-server to v0.5.0.

dgrisonnet:
Create the pod lister based on a filtered informer factory that will
filter non-running pods so that prometheus-adapter don't expect metrics
from them.

Fixes #402

### DIFF
--- a/cmd/adapter/adapter.go
+++ b/cmd/adapter/adapter.go
@@ -30,8 +30,12 @@ import (
 	basecmd "github.com/kubernetes-sigs/custom-metrics-apiserver/pkg/cmd"
 	"github.com/kubernetes-sigs/custom-metrics-apiserver/pkg/provider"
 
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


s-urbaniak:
great catch, thank you! :tada:

PR #401, opened 2021-05-25
Remove vendor directory

dgrisonnet:
/cc @s-urbaniak @fpetkovski

PR #406, opened 2021-06-02
Makefile: consolidate docker-build

dgrisonnet:
As part of this PR, I upgraded the golang image used for building to
1.16 and consolidated how the docker-build rule was working. Previously,
it was failing in master's CI [1] because the go modules were not downloaded
in the build image. To improve that, I replaced the combination of
docker run and docker build by a multi-stage Dockerfile responsible for
building the adapter and running it.

In addition to that, I removed the `_output` directory completely as it
wasn't really meaningful to have it anymore. I also removed the
`build-local-image` rule as it was a duplicate of the `docker-build`
rule with the only difference of using a scratch base image.

Also, since all the base images that we are using by default are based
on busybox, I change the UID used in the image to 65534 which correspond
to the nobody user in busybox.

[1] https://travis-ci.org/github/kubernetes-sigs/prometheus-adapter/builds/773263282

Recent CI failures on merge commits: https://github.com/kubernetes-sigs/prometheus-adapter/commits/master after #401 

/cc @s-urbaniak @fpetkovski


fpetkovski (commented):
lgtm

PR #431, opened 2021-07-13
Prevent prometheus-adapter from returning negative resource metrics

dgrisonnet:
When serving the resource metrics API, prometheus-adapter may return
negative values for pods/nodes memory and CPU usage. This happens
because Prometheus sees counter resets which results in Prometheus
interpolating data incorrectly to avoid the counter value going down.
To prevent that, we need to add some guards in prometheus-adapter to
replace the negative value with zero whenever it detects one.

/cc @s-urbaniak @fpetkovski

### DIFF
--- a/pkg/resourceprovider/provider.go
+++ b/pkg/resourceprovider/provider.go
@@ -19,6 +19,7 @@ package resourceprovider
 import (
 	"context"
 	"fmt"
+	"math"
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
:+1:

fpetkovski:
lgtm

PR #432, opened 2021-07-13
Support setting headers on requests to Prometheus

discordianfish:


cmd/adapter/adapter_test.go
```diff
@@ -146,3 +147,41 @@ func TestMakePrometheusCAClient(t *testing.T) {
 		}
 	}
 }
+
+func TestParseHeaderArgs(t *testing.T) {
+
```
fpetkovski, line 152:
```suggestion
```

fpetkovski (approved):
Just one small nit, otherwise looks great! Thanks for the contribution 👍

cmd/adapter/adapter_test.go
discordianfish: I've had this in there to be consistent with other functions here. Fixed

fpetkovski: Ah, I wasn't aware of that. Thanks for fixing it nevertheless :)

cmd/adapter/adapter.go
```diff
@@ -406,3 +410,16 @@ func makePrometheusCAClient(caFilePath string, tlsCertFilePath string, tlsKeyFil
 		},
 	}, nil
 }
+
+func parseHeaderArgs(args []string) map[string][]string {
```
dgrisonnet, line 414: Can we use `http.Header` instead of `map[string][]string`? We could then add the headers via the `Add()` function of http.Headers.

cmd/adapter/adapter.go
```diff
@@ -406,3 +410,16 @@ func makePrometheusCAClient(caFilePath string, tlsCertFilePath string, tlsKeyFil
...
+	headers := make(map[string][]string, len(args))
+	for _, h := range args {
+		parts := strings.SplitN(h, "=", 2)
+		value := ""
+		if len(parts) > 1 {
```
dgrisonnet, line 419: else sounds like a misconfiguration that we should make users aware of by returning an error here

pkg/client/api.go
```diff
@@ -47,6 +47,7 @@ type GenericAPIClient interface {
 type httpAPIClient struct {
 	client  *http.Client
 	baseURL *url.URL
+	headers map[string][]string
```
dgrisonnet, line 50: same here, can we use `http.Header` type instead?

dgrisonnet (changes requested):
Thank you @discordianfish, that will be a great addition to the project. Just a couple of comments from my side, but it looks good overall.

cmd/adapter/adapter.go
discordianfish: Thought about that but it looks like headers can have no value, e.g: https://github.com/whatwg/fetch/issues/332 - probably not useful in this context though so open either way.
So, should I return an error?

discordianfish:
@dgrisonnet PTAL

cmd/adapter/adapter.go
dgrisonnet: Oh, good catch! I double-checked the RFC and empty headers are indeed valid. Since that's the case, I am fine with not erroring here.

pkg/client/api.go
```diff
@@ -113,10 +115,11 @@ func (c *httpAPIClient) Do(ctx context.Context, verb, endpoint string, query url
 }
 
 // NewGenericAPIClient builds a new generic Prometheus API client for the given base URL and HTTP Client.
-func NewGenericAPIClient(client *http.Client, baseURL *url.URL) GenericAPIClient {
+func NewGenericAPIClient(client *http.Client, baseURL *url.URL, headers map[string][]string) GenericAPIClient {
```
dgrisonnet, line 118: we should also use `http.Header` type here

pkg/client/api.go
```diff
@@ -139,8 +142,8 @@ func NewClientForAPI(client GenericAPIClient) Client {
 }
 
 // NewClient creates a Client for the given HTTP client and base URL (the location of the Prometheus server).
-func NewClient(client *http.Client, baseURL *url.URL) Client {
-	genericClient := NewGenericAPIClient(client, baseURL)
+func NewClient(client *http.Client, baseURL *url.URL, headers map[string][]string) Client {
```
dgrisonnet, line 145: ditto

dgrisonnet (approved):
Awesome! Thank you @discordianfish :)

/ok-to-test
/lgtm
/approve

Issue #369, opened 2021-02-08
Prometheus-adapter isn't serving OpenAPI spec for resource metrics


dgrisonnet:
When running prometheus-adapter v0.8.3 and trying to run `kubectl explain`  on PodMetrics and NodeMetrics, the commands fail with an error message saying that the resource couldn't be found.
```
$ kubectl explain PodMetrics
error: couldn't find resource for "metrics.k8s.io/v1beta1, Kind=PodMetrics"
$ kubectl explain NodeMetrics
error: couldn't find resource for "metrics.k8s.io/v1beta1, Kind=NodeMetrics"
```
I would have expected to get the OpenAPI spec for the queried resource. For instance, trying to do the same with metrics-server v0.4.1 gives us the expected result.
```
$ kubectl explain PodMetrics
KIND:     PodMetrics
VERSION:  metrics.k8s.io/v1beta1

DESCRIPTION:
     PodMetrics sets resource usage metrics of a pod.

FIELDS:
   apiVersion	<string>
     APIVersion defines the versioned schema of this representation of an
     object. Servers should convert recognized schemas to the latest internal
     value, and may reject unrecognized values. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#resources

   containers	<[]Object> -required-
     Metrics for all containers are collected within the same time window.

   kind	<string>
     Kind is a string value representing the REST resource this object
     represents. Servers may infer this from the endpoint the client submits
     requests to. Cannot be updated. In CamelCase. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds

   metadata	<Object>

   timestamp	<string> -required-
     The following fields define time interval from which metrics were collected
     from the interval [Timestamp-Window, Timestamp].

   window	<string> -required-
```

PR #435, opened 2021-07-15
fix: incorrect type used for openapi spec

arajkumar:
Prior to this fix, openapi spec for prometheus-adapter apiextension was based on the type "k8s.io/sample-apiserver/pkg/apiserver" which is incorrect. Due to the incorrect type, `kubectl explain podmetrics` (or nodemetrics) wasn't showing any doc for any resources from metrics.k8s.io/v1beta1.

This changeset fixes the problem by using the right type(sigs.k8s.io/metrics-server/pkg/api) for the openapi generation.

This also helped to remove the sample-apiserver dependency from
prometheus-adapter.

Fixes https://github.com/kubernetes-sigs/prometheus-adapter/issues/369

Signed-off-by: Arunprasad Rajkumar <arajkuma@redhat.com>


arajkumar:
I signed it

dgrisonnet (approved):
Oh wow, good catch @arajkumar, and congratz on your first PR to the kubernetes ecosystem! :tada: :rocket: 

I guess we will have the same issue with custom and external metrics since we are only using the Scheme from metrics-server, but I wouldn't block this fix because of that.
Ultimately what we need to do is to dependency inject the Scheme instead of using a global Scheme everywhere as in https://github.com/kubernetes-sigs/custom-metrics-apiserver/blob/master/pkg/apiserver/apiserver.go#L34-L59. I believe metrics-server is also doing the same so it might be worth fixing it in both projects.

/ok-to-test
/lgtm
/approve

arajkumar:
Thanks @dgrisonnet for the quick review. I will look at the rest of the problems you have mentioned and address those as well.

PR #436, opened 2021-07-16
fix: add openapi spec for custom and external metrics types

arajkumar:
Signed-off-by: Arunprasad Rajkumar <arajkuma@redhat.com>

### DIFF
--- a/cmd/adapter/adapter.go
+++ b/cmd/adapter/adapter.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"time"
 
+	customexternalmetrics "github.com/kubernetes-sigs/custom-metrics-apiserver/pkg/apiserver"
 	basecmd "github.com/kubernetes-sigs/custom-metrics-apiserver/pkg/cmd"
 	"github.com/kubernetes-sigs/custom-metrics-apiserver/pkg/provider"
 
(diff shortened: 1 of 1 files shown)


cmd/adapter/adapter.go
```diff
@@ -30,6 +30,7 @@ import (
 	basecmd "github.com/kubernetes-sigs/custom-metrics-apiserver/pkg/cmd"
 	"github.com/kubernetes-sigs/custom-metrics-apiserver/pkg/provider"
 
+	customexternalmetrics "github.com/kubernetes-sigs/custom-metrics-apiserver/pkg/apiserver"
```
dgrisonnet, line 33: this import should be moved next to the other `custom-metrics-apiserver` imports

arajkumar: Thanks @dgrisonnet. Fixed it.

cmd/adapter/adapter.go
```diff
@@ -41,7 +42,6 @@ import (
 	"k8s.io/client-go/transport"
 	"k8s.io/component-base/logs"
 	"k8s.io/klog/v2"
-
```
dgrisonnet, line 44: let's keep this line break, it makes the imports more readable

arajkumar: Ack.

Issue #425, opened 2021-07-08
Decouple /metrics endpoint from generic apiserver


dgrisonnet:
Prometheus-adapter /metrics endpoint exposes metrics about the adapter health and these metrics are currently entirely coming from the generic apiserver implementation which installs the metrics handler on the same server as the one used to serve the metrics API to the apiserver.
This becomes problematic when we want to add authz/authn to metrics scraping since one would have to be authorized by the apiserver in order to access prometheus-adapter's server as well as having the permission to get the /metrics endpoint. Whereas if we were to decouple the /metrics endpoint from the generic API server, the scraper would only have to be authorized to scrape metrics in order to get prometheus-adapter metrics.

If we were to add authz/authn today without decoupling, we would have to add a kube-rbac-proxy in front of the `/metrics` and point the scraper to the proxy. It works fine when the request goes from the scraper to the proxy, but from the proxy to the `/metrics` endpoint, the proxy needs to forward the bearer token used by the apiserver in order to be authorized and this is not in the scope of kube-rbac-proxy which is expected to forward requests to applications without kubernetes awareness.

To decouple the /metrics endpoint from the generic apiserver, there is a field called [EnableMetrics](https://github.com/kubernetes/apiserver/blob/v0.21.2/pkg/server/config.go#L124) in the apiserver config to prevent it from creating a metrics handler. Then, we would need to bring our own registry and server to serve the metrics that we previously added by the apiserver.

FWIW, there was a similar effort in metrics server, and I would expect us to do something similar: https://github.com/kubernetes-sigs/metrics-server/pull/587.

dgrisonnet:
cc @prashbnair

dgrisonnet:
To solve this issue we can set `--authorization-always-allow-paths=/metrics` to disable authorization on the metrics endpoint.

PR #451, opened 2021-09-13
Decouple /metrics endpoint from generic apiserver

prashbnair:
This PR is a fix for this [issue](https://github.com/kubernetes-sigs/prometheus-adapter/issues/425)


prashbnair:
I signed it

fpetkovski:
@prashbnair you can do 
```
git commit --amend --no-edit --signoff
```
followed by a force push to add the signature.

cmd/adapter/adapter.go
```diff
@@ -334,12 +335,27 @@ func main() {
...
 	}
 }
 
+// Disables the metrics endpoint in the api server and brings up a new metrics endpoint
+func setupMetricsServer(config *customexternalmetrics.Config) {
+	config.GenericConfig.EnableMetrics = false
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", legacyregistry.Handler())
+
+	go http.ListenAndServe(":9090", mux)
```
dgrisonnet, line 356: The server address and port should be configurable if we want to be consistent with what was previously available. Also, since the generic apiserver was serving the metrics over HTTPS, we should do the same with the new server.

cmd/adapter/adapter.go
```diff
@@ -334,12 +335,27 @@ func main() {
...
+
 	// run the server
 	if err := cmd.Run(stopCh); err != nil {
 		klog.Fatalf("unable to run custom metrics adapter: %v", err)
 	}
 }
 
+// Disables the metrics endpoint in the api server and brings up a new metrics endpoint
+func setupMetricsServer(config *customexternalmetrics.Config) {
+	config.GenericConfig.EnableMetrics = false
```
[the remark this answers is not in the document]
prashbnair: Done

cmd/adapter/adapter.go
prashbnair: Done

cmd/adapter/adapter.go
```diff
@@ -144,6 +161,12 @@ func (cmd *PrometheusAdapter) addFlags() {
 		"interval at which to re-list the set of all available metrics from Prometheus")
 	cmd.Flags().DurationVar(&cmd.MetricsMaxAge, "metrics-max-age", cmd.MetricsMaxAge, ""+
 		"period for which to query the set of available metrics from Prometheus")
+	cmd.Flags().StringVar(&cmd.secureMetricsPort, "secure-metrics-port", cmd.secureMetricsPort,
+		"Secure port for metrics")
+	cmd.Flags().StringVar(&cmd.metricsHost, "metrics-host", cmd.metricsHost,
+		"Host for exposing metrics")
```
dgrisonnet, line 167: Could we add some defaults?

cmd/adapter/adapter.go
```diff
@@ -144,6 +161,12 @@ func (cmd *PrometheusAdapter) addFlags() {
 		"interval at which to re-list the set of all available metrics from Prometheus")
 	cmd.Flags().DurationVar(&cmd.MetricsMaxAge, "metrics-max-age", cmd.MetricsMaxAge, ""+
 		"period for which to query the set of available metrics from Prometheus")
+	cmd.Flags().StringVar(&cmd.secureMetricsPort, "secure-metrics-port", cmd.secureMetricsPort,
```
dgrisonnet, line 164: Users shouldn't be only able to configure secure ports, they should be able to choose whatever fits their needs. As such I wouldn't mention the `secure` part in the option name.

deploy/manifests/custom-metrics-apiserver-deployment.yaml
```diff
@@ -49,3 +58,9 @@ spec:
           name: adapter-config
       - name: tmp-vol
         emptyDir: {}
+      - configMap:
+          name: metrics-tls-config
+        name: metricstlsconfig
```
dgrisonnet, line 63: Nit: keeping the `-` would make the name more readible

deploy/manifests/custom-metrics-apiserver-deployment.yaml
```diff
@@ -40,6 +43,12 @@ spec:
           readOnly: true
         - mountPath: /tmp
           name: tmp-vol
+        - mountPath: /etc/tls
+          name: metricstlsconfig
+          readOnly: false
+        - mountPath: /etc/certs
+          name: servingcerts
+          readOnly: false
```
dgrisonnet, line 51: we shouldn't need these certs

cmd/adapter/adapter.go
```diff
@@ -144,6 +161,9 @@ func (cmd *PrometheusAdapter) addFlags() {
 		"interval at which to re-list the set of all available metrics from Prometheus")
 	cmd.Flags().DurationVar(&cmd.MetricsMaxAge, "metrics-max-age", cmd.MetricsMaxAge, ""+
 		"period for which to query the set of available metrics from Prometheus")
+	cmd.Flags().StringVar(&cmd.MetricsPort, "metrics-port", cmd.MetricsPort, "Port for metrics")
+	cmd.Flags().StringVar(&cmd.MetricsHost, "metrics-host", cmd.MetricsHost, "Host for exposing metrics")
```
dgrisonnet, line 165: Could we perhaps also mention the default value here? For example:
```suggestion
	cmd.Flags().StringVar(&cmd.MetricsPort, "metrics-port", cmd.MetricsPort, "Port to expose prometheus-adapter self metrics on. (default: 8080)")
	cmd.Flags().StringVar(&cmd.MetricsHost, "metrics-host", cmd.MetricsHost, "Host to expose prometheus-adapter self metrics on. (default: "0.0.0.0")")
```

cmd/adapter/adapter.go
```diff
@@ -144,6 +161,9 @@ func (cmd *PrometheusAdapter) addFlags() {
...
+	cmd.Flags().StringVar(&cmd.TLSConfig, "tls-config", cmd.TLSConfig, "TLS configuration for metrics server")
```
dgrisonnet, line 166: Could please mention that the tls config is optional as it is done for the other TLS related flags.

cmd/adapter/adapter.go
```diff
@@ -275,6 +295,8 @@ func main() {
 	cmd := &PrometheusAdapter{
 		PrometheusURL:         "https://localhost",
 		MetricsRelistInterval: 10 * time.Minute,
+		MetricsHost:           "0.0.0.0",
```
dgrisonnet, line 298: The `::` wildcard might be a better default since it would work for both ipv4 and ipv6.

cmd/adapter/adapter.go
```diff
@@ -334,12 +356,35 @@ func main() {
 		klog.Fatalf("unable to install resource metrics API: %v", err)
 	}
 
+	config, err := cmd.Config()
+	if err != nil {
+		klog.Fatalf("unable to get configuration of custom metrics apiserver: %v", err)
+	}
+
+	apiServerConfig(config)
+	cmd.setupMetricsServer()
```
dgrisonnet, line 365: To be able to shutdown both servers in case anything happens to one of them, it would be better to have them in run groups. You can take this implementation from kube-state-metrics as an example: https://github.com/kubernetes/kube-state-metrics/blob/master/main.go#L182-L203

cmd/adapter/adapter.go
```diff
@@ -334,12 +356,35 @@ func main() {
...
 	if err := cmd.Run(stopCh); err != nil {
 		klog.Fatalf("unable to run custom metrics adapter: %v", err)
 	}
 }
 
+func apiServerConfig(config *customexternalmetrics.Config) {
+	config.GenericConfig.EnableMetrics = false
+}
+
+func (cmd PrometheusAdapter) setupMetricsServer() {
```
dgrisonnet, line 377: it would be better to take the PrometheusAdapter by pointer here

cmd/adapter/adapter.go
```diff
@@ -81,6 +95,9 @@ type PrometheusAdapter struct {
 	// MetricsMaxAge is the period to query available metrics for
 	MetricsMaxAge time.Duration
 
+	MetricsPort string
+	TLSConfig   string
+	MetricsHost string
```
dgrisonnet, line 100: could you perhaps document these fields as done above?

deploy/manifests/custom-metrics-apiserver-deployment.yaml
dgrisonnet: By default, I don't think we should add certs since ultimately they will have to be managed by the users so they will have to add their own TLS config if they want to setup the server properly. In the meantime we can have the metrics exposed over HTTPS but we can skip the cert verification by setting `insecureSkipTLSVerify: true`.

cmd/adapter/adapter.go
prashbnair: Isnt this line 
https://github.com/kubernetes-sigs/prometheus-adapter/pull/451/files#diff-4d0b81c92e53ab757dd6ce4dca5fdf332665e5e6d01e2e8465e517f5549886cdR385 effectively doing the same thing?  In this case I do not have start up of the other server as it is in the custom-metrics-apiserver

deploy/manifests/custom-metrics-apiserver-deployment.yaml
prashbnair: Done

cmd/adapter/adapter.go
prashbnair: Done

cmd/adapter/adapter.go
prashbnair: Done

cmd/adapter/adapter.go
prashbnair: Done

cmd/adapter/adapter.go
prashbnair: I have implemented the run group for just one of them, the second one does accept a context. Let me know if this is useful, otherwise will revert.

cmd/adapter/adapter.go
prashbnair: Done

cmd/adapter/adapter.go
prashbnair: Done

deploy/manifests/custom-metrics-apiserver-deployment.yaml
prashbnair: The insecureSkipTLSVerify: true is already present at the ApiService level.

dgrisonnet:
It was pointed out on metrics-server that we could use `--authorization-always-allow-paths=/metrics` to disable authorization on the metrics endpoint which should solve the problem that we have. I completely missed this flag in the first place, sorry about that. Do you think there is still a use-case for this PR or should we close it in favor of using the flag?

PR #453, opened 2021-09-15
Update walkthrough.md (fix selector)

reneluria:
selecting label with `metricLabelSelector` and not `selector` while calling the API
cf. https://github.com/kubernetes-sigs/custom-metrics-apiserver/blob/0ca2b1909cdc45259654a4e06cac40f98fa1bcb0/pkg/apiserver/installer/conversion.go#L36

### DIFF
--- a/docs/walkthrough.md
+++ b/docs/walkthrough.md
@@ -175,7 +175,7 @@ spec:
 If you try creating that now (and take a look at your controller-manager
 logs), you'll see that the that the HorizontalPodAutoscaler controller is
 attempting to fetch metrics from
-`/apis/custom.metrics.k8s.io/v1beta1/namespaces/default/pods/*/http_requests?selector=app%3Dsample-app`,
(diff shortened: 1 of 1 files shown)


docs/walkthrough.md
```diff
@@ -175,7 +175,7 @@ spec:
 If you try creating that now (and take a look at your controller-manager
 logs), you'll see that the that the HorizontalPodAutoscaler controller is
 attempting to fetch metrics from
-`/apis/custom.metrics.k8s.io/v1beta1/namespaces/default/pods/*/http_requests?selector=app%3Dsample-app`,
+`/apis/custom.metrics.k8s.io/v1beta1/namespaces/default/pods/*/http_requests?metricLabelSelector=app%3Dsample-app`,
```
dgrisonnet, line 178: The selector should be `labelSelector` instead of `metricLabelSelector` since `app=sample-app` is a label of the pod, not of the metric.

reneluria: Ok, nevermind

PR #455, opened 2021-09-24
Fix external metrics provider not respecting metrics-max-age

leoskyrocker:

PR #457, opened 2021-09-28
fix: add support for maxAge in external provider

jerome-quere:
Hello,

Not sure if this is correct but the `--maxAge` flag is no taken into account for the `external` metrics provider.


jerome-quere:
Oops it's a duplicate of #455 I'm closing this one.

PR #499, opened 2022-05-01
Set non zero query parameter for end in series query API

srinidhis94:
**Current behavior**
* When we query for series metric, we pass zero as value for end query key.
Ex : http://localhost:9090/api/v1/series?match[]=my_metric&start=1651129852&end=0'
* The error is similar to [this](https://github.com/prometheus/prometheus/issues/4197).
* While this works on self hosted prometheus, I faced an issue where seriesQuery started throwing error on managed prometheus provided by GCP.
* It worked after I started using custom build passing valid end time.

### DIFF
--- a/pkg/custom-provider/provider.go
+++ b/pkg/custom-provider/provider.go
@@ -238,7 +238,7 @@ type selectorSeries struct {
 
 func (l *cachingMetricsLister) updateMetrics() error {
 	startTime := pmodel.Now().Add(-1 * l.maxAge)
-
+	endTime := pmodel.Now()
 	// don't do duplicate queries when it's just the matchers that change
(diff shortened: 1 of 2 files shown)


dgrisonnet:
Thank you for the contribution @srinidhis94 :)

/lgtm
/approve
/ok-to-test

dgrisonnet:
@srinidhis94 will you have some time to finish this work? It seems to be waiting to be rebased and the tests need to be fixed

srinidhis94:
Hey I'll check this weekend .

olivierlemasle:
Hi @srinidhis94
Do you think you'll be able to rebase this PR?

PR #509, opened 2022-06-28
Add --prometheus-verb to support POST requests to prometheus servers

ksauzz:
We hit the limitation of HTTP request line size of a certain managed prometheus service. To avoid the limitation, we would like to use POST method to query Prometheus servers.

Docs: https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries

> You can URL-encode these parameters directly in the request body by using the POST method and Content-Type: application/x-www-form-urlencoded header. This is useful when specifying a large query that may breach server-side URL character limits.

### DIFF
--- a/cmd/adapter/adapter.go
+++ b/cmd/adapter/adapter.go
@@ -93,7 +93,7 @@ func (cmd *PrometheusAdapter) makePromClient() (prom.Client, error) {
 	}
 
 	if cmd.PrometheusVerb != http.MethodGet && cmd.PrometheusVerb != http.MethodPost {
-		return nil, fmt.Errorf("unsupported Prometheus Http Verb %q", cmd.PrometheusVerb)
+		return nil, fmt.Errorf("unsupported Prometheus HTTP verb %q. use \"GET\" or \"POST\" instead.", cmd.PrometheusVerb)
 	}
 
 	var httpClient *http.Client
@@ -144,7 +144,7 @@ func (cmd *PrometheusAdapter) addFlags() {
 	cmd.Flags().StringArrayVar(&cmd.PrometheusHeaders, "prometheus-header", cmd.PrometheusHeaders,
 		"Optional header to set on requests to prometheus-url. Can be repeated")
 	cmd.Flags().StringVar(&cmd.PrometheusVerb, "prometheus-verb", cmd.PrometheusVerb,
-		"HTTP Verb to set on requests to Prometheus.")
+		"HTTP verb to set on requests to Prometheus. Possible values: \"GET\", \"POST\"")
 	cmd.Flags().StringVar(&cmd.AdapterConfigFile, "config", cmd.AdapterConfigFile,
 		"Configuration file containing details of how to transform between Prometheus metrics "+
 			"and custom metrics API resources")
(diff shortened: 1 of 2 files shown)


ksauzz:
/assign @dgrisonnet
thanks

cmd/adapter/adapter.go
```diff
@@ -137,6 +143,8 @@ func (cmd *PrometheusAdapter) addFlags() {
 		"Optional file containing the bearer token to use when connecting with Prometheus")
 	cmd.Flags().StringArrayVar(&cmd.PrometheusHeaders, "prometheus-header", cmd.PrometheusHeaders,
 		"Optional header to set on requests to prometheus-url. Can be repeated")
+	cmd.Flags().StringVar(&cmd.PrometheusVerb, "prometheus-verb", cmd.PrometheusVerb,
+		"HTTP Verb to set on requests to Prometheus.")
```
dgrisonnet, line 147: Could you specify the default value as well as the possible values for the verbs?

cmd/adapter/adapter.go
```diff
@@ -90,6 +92,10 @@ func (cmd *PrometheusAdapter) makePromClient() (prom.Client, error) {
 		return nil, fmt.Errorf("invalid Prometheus URL %q: %v", baseURL, err)
 	}
 
+	if cmd.PrometheusVerb != http.MethodGet && cmd.PrometheusVerb != http.MethodPost {
+		return nil, fmt.Errorf("unsupported Prometheus Http Verb %q", cmd.PrometheusVerb)
```
dgrisonnet, line 96:
```suggestion
		return nil, fmt.Errorf("unsupported Prometheus HTTP verb %q", cmd.PrometheusVerb)
```
Could you also add the list of possible verb value to this error to make it simpler for the users?

pkg/client/api.go
```diff
@@ -53,13 +54,26 @@ type httpAPIClient struct {
...
 	if err != nil {
 		return APIResponse{}, fmt.Errorf("error constructing HTTP request to Prometheus: %v", err)
 	}
 	req.WithContext(ctx)
-	req.Header = c.headers
+	for key, values := range c.headers {
+		for _, value := range values {
+			req.Header.Add(key, value)
+		}
+	}
```
dgrisonnet, line 73: what is the reasoning behind this change? It shouldn't be needed

dgrisonnet (commented):
Made a few comments but it looks good to me overall

dgrisonnet:
I will be cutting a new release very soon, so we might be able to get this change in.

pkg/client/api.go
ksauzz: In my understanding, the original code set the reference of the singleton http headers object to `req.Header`, then `req.Header.Set("Content-Type", "application/x-www-form-urlencoded")` modified the singleton object. Therefore all subsequent http reqeust has `Content-Type: application/x-www-form-urlencoded` regardless of the http method.
Is the behavior no problem?

ksauzz:
Thank you for the review! I pushed the fixes per the review.

### New help message

```
% ./adapter --help 2>&1| grep prometheus-verb              
      --prometheus-verb string                                  HTTP verb to set on requests to Prometheus. Possible values: "GET", "POST" (default "GET")
```
### New error message

```
% ./adapter --prometheus-verb=DELETE 2>&1| grep unsupported
F0812 13:18:06.485853  311844 adapter.go:308] unable to construct Prometheus client: unsupported Prometheus HTTP verb "DELETE". use "GET" or "POST" instead.

```

dgrisonnet:
Awesome! Thank you for looking into this @ksauzz. I will add this new feature to the upcoming release.

/lgtm
/approve

ksauzz:
Thank you for your help!!!

PR #491, opened 2022-03-10
Update README.md with Helm-3 Command

Ruwan-Ranganath:
Updated for Helm3 command removed "--name"

PR #510, opened 2022-07-01
docs(readme): fix installation command for helm3

sumo-drosiek:
Helm3 install doesn't have `--name` flag

`Error: unknown flag: --name`


sumo-drosiek:
Closing in favor of #491

Issue #426, opened 2021-07-08
Update deployment manifests and instructions


dgrisonnet:
Deployment manifests and instructions are very outdated in the project since the community has been solely updating the kube-prometheus manifests. It would be great to leverage the efforts that were made there and update the manifests that we have in this repository.

championshuttler:
@dgrisonnet, can you help me with getting started what i need to do? thanks

dgrisonnet:
Sure, so the goal of this issue is to update the deployment manifests and instructions that we have in this repository under the [deploy](https://github.com/kubernetes-sigs/prometheus-adapter/tree/master/deploy) directory based on the more up-to-date manifests in kube-prometheus that are prefixed by `prometheus-adapter-` in the [manifests](https://github.com/prometheus-operator/kube-prometheus/tree/main/manifests) directory.

The few pointers that I could give you are that you should be able to update all the manifests and bring the new ones except for the [configmap](https://github.com/kubernetes-sigs/prometheus-adapter/blob/89425b72ccd2124af3ed898bd07bf6eee50e02a5/deploy/manifests/custom-metrics-config-map.yaml) and the [apiservices](https://github.com/kubernetes-sigs/prometheus-adapter/blob/89425b72ccd2124af3ed898bd07bf6eee50e02a5/deploy/manifests/custom-metrics-apiservice.yaml) that we can keep as is, except for a few selectors that might need to be updated to match the new changes.
Also, it would be great to move away from the `custom-metrics-apiserver` resource name to `prometheus-adapter` which is used in kube-prometheus and is in my opinion more clear.
As for the deployment [instructions](https://github.com/kubernetes-sigs/prometheus-adapter/tree/89425b72ccd2124af3ed898bd07bf6eee50e02a5/deploy#example-deployment), we could move the second point out of the default deployment and add a new section explaining how users can configure the adapter and enable HTTPS.

Let me know if you need any additional information and feel free to reach out to me on slack if you need any help :slightly_smiling_face:

dgrisonnet:
This was fixed by https://github.com/kubernetes-sigs/prometheus-adapter/pull/531.

PR #516, opened 2022-07-18
Use node_exporter metrics for node resources

SuperQ:
Use the metrics provided by the node_exporter for node-level resources.

This is much more efficient, especially for memory, as it's only two
metrics per node, rather than based on the number of running containers.

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/cmd/config-gen/utils/default.go
+++ b/cmd/config-gen/utils/default.go
@@ -94,7 +94,7 @@ func DefaultConfig(rateInterval time.Duration, labelPrefix string) *config.Metri
 		ResourceRules: &config.ResourceRules{
 			CPU: config.ResourceRule{
 				ContainerQuery: fmt.Sprintf("sum(rate(container_cpu_usage_seconds_total{<<.LabelMatchers>>}[%s])) by (<<.GroupBy>>)", pmodel.Duration(rateInterval).String()),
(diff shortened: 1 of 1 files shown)


dgrisonnet:
This is intended to avoid having a dependency on node-exporter. The basic configuration provided by this repository should be installable on any Kubernetes cluster with a Prometheus server running.

For the general use case, I totally agree that node-exporter is better suited than CRI metrics, but I don't think it should be the default choice. We could have multiple configurations in this repository but there is a more general configuration overhaul to do in the project before that https://github.com/kubernetes-sigs/prometheus-adapter/issues/426.

Since I see that you tagged this PR in a kube-prometheus-stack issue, they could do the same as kube-prometheus and just write their own configuration instead of reusing the outdated one from this project. Kube-prometheus has the configuration that is the most up-to-date since it was written by the prometheus-adapter maintainer in favor of the configuration in this project, so maybe they could base the helm config on it. https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/prometheusAdapter-configMap.yaml

dgrisonnet:
@SuperQ it was recently brought to my attention that we can't get node-level information from the kubelet `/metrics/resource` endpoint, we would have to rely on either node-exporter metrics, Prometheus k8s discovery node label or the worst option sticking with /metrics/cadvisor. As such I think that adding a dependency on node-exporter is the best option and would be interested in reviving this effort. Could you perhaps rebase the PR?

dgrisonnet:
I got confused and forgot that we could rely on `node_cpu_usage_seconds_total` and `node_memory_working_set_bytes` that are coming from kubelet. Going forward that is what we will be going and it should fix the issues reported against the chart. Sorry about that.

PR #519, opened 2022-07-28
Patch upgrade Golang from 1.16.4 to 1.16.15

gburton1:
There are a large number of CVEs against prometheus-adapter, and most of them are related to running behind on patches of Golang. This will solve many of them.

Signed-off-by: Greg Burton <9094087+gburton1@users.noreply.github.com>


dgrisonnet:
Hi @gburton1, do you specifically need this version of Golang or would you be also fine with 1.18.5? I am considering doing a complete dependency update before cutting the new release that would contain that change.

VERSION
```diff
@@ -1 +1 @@
-0.9.1
+0.9.2
```
dgrisonnet, line 1: this is reserved for when we cut a new release of the project and since this PR is just meant to upgrade the golang version, could you please revert this change?

gburton1:
Any supported Golang major line is fine because they should all offer the same security patches. 1.18.5 works fine!

gburton1:
Reverted the VERSION bump. If you want to handle the jump to 1.18.5 in another PR yourself, since it might involve other changes, I can close this one. But if you expect that to take some time, I would appreciate this trivial patch bump + a new release to get us past a ton of CVEs!

dgrisonnet:
Sure let me do that right now

gburton1:
@dgrisonnet ping me when you have an alternative PR/release. Once we have a new release published with the Golang upgrade, I'll go shop it around to the maintainers of the Prometheus Adapter Helm chart to get a new release there also. This is my path to closing many CVEs in our system against the prom adapter!

PR #531, opened 2022-09-02
Updates deploy/manifest to latest version in sync with kube-prom

JoaoBraveCoding:
Issue https://github.com/kubernetes-sigs/prometheus-adapter/issues/426


deploy/manifests/api-service.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: apiregistration.k8s.io/v1
+kind: APIService
+metadata:
+  labels:
+    app.kubernetes.io/component: metrics-adapter
+    app.kubernetes.io/name: prometheus-adapter
+    app.kubernetes.io/version: 0.9.1
```
dgrisonnet, line 7: v0.10.0 was released quite recently, would you mind updating it in kube-prometheus and regenerate the manifests so that the manifests that we provide in prometheus-adapter target its latest release? IIRC you would just have to update the value in https://github.com/prometheus-operator/kube-prometheus/blob/main/jsonnet/kube-prometheus/versions.json#L8 and regenerate the manifests.

deploy/manifests/config-map.yaml
```diff
@@ -0,0 +1,68 @@
...
+    "resourceRules":
+      "cpu":
+        "containerLabel": "container"
+        "containerQuery": |
+          sum by (<<.GroupBy>>) (
+            irate (
+                container_cpu_usage_seconds_total{<<.LabelMatchers>>,container!="",pod!=""}[4m]
+            )
+          )
+        "nodeQuery": |
```
dgrisonnet, line 13: For node queries, I would prefer if we stick to kubernetes metrics as kubernetes should be the only dependency of the default installation of prometheus-adapter. So instead of using metrics from node-exporter and windows-exporter, we could reuse the container queries but with `id='/'` as an additional label selector to get the resource usage of the node.

In the future, we could make the different setup configurable via jsonnet, but for now, let's just stick to the default use-case.

dgrisonnet (commented):
We should also consider updating the README and the Walkthrough, they still have references to the old manifests.

Also, something we could consider doing in the future would be to set up a simple CI job that would check that the manifests are getting applied in a cluster properly and that prometheus-adapter is running. Ideally we should do that during the e2e tests but since we don't have any, just a simple check verifying that the pod is running would be a great start.

JoaoBraveCoding:
@dgrisonnet while upgrading de Walkthrough, I modified the version of the HPA created to v2beta2 since v2beta1 is unavailable on 1.25+. However, in v2beta2 if we configure the HPA to scale on pod metrics then it will look for pod metrics under the API `pod.custom.metrics.k8s.io` while v2beta1 looks for them under `custom.metrics.k8s.io`. Does this mean that in 1.25+ if we want to HPA with custom metrics using the adapter then we will have to register an APIService that will service metrics under `pod.custom.metrics.k8s.io` instead of the usual `custom.metrics.k8s.io`, is this the correct way of doing things or am I missing something?

JoaoBraveCoding:
> @dgrisonnet while upgrading de Walkthrough, I modified the version of the HPA created to v2beta2 since v2beta1 is unavailable on 1.25+. However, in v2beta2 if we configure the HPA to scale on pod metrics then it will look for pod metrics under the API `pod.custom.metrics.k8s.io` while v2beta1 looks for them under `custom.metrics.k8s.io`. Does this mean that in 1.25+ if we want to HPA with custom metrics using the adapter then we will have to register an APIService that will service metrics under `pod.custom.metrics.k8s.io` instead of the usual `custom.metrics.k8s.io`, is this the correct way of doing things or am I missing something?

I think I got it working, will submit an update to the PR soon

JoaoBraveCoding:
> Also, something we could consider doing in the future would be to set up a simple CI job that would check that the manifests are getting applied in a cluster properly and that prometheus-adapter is running. Ideally, we should do that during the e2e tests but since we don't have any, just a simple check verifying that the pod is running would be a great start.

100% agree, I could also help with that but I would prefer to do it in a follow-up PR if it's OK, if you want I can create an issue for it.

> In the future, we could make the different setup configurable via jsonnet, but for now, let's just stick to the default use-case.

Also agree that this would be nice, I'm happy to also help with it in a followup

deploy/manifests/service-monitor.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
```
dgrisonnet, line 2: we should add the ServiceMonitor manifest here since that would add a dependency on prometheus-operator

dgrisonnet (commented):
One last comment and we are good to go :tada:

deploy/manifests/config-map.yaml
```diff
@@ -0,0 +1,53 @@
...
+          sum by (<<.GroupBy>>) (
+            container_memory_working_set_bytes{<<.LabelMatchers>>,container!="",pod!=""}
+          )
+        "nodeQuery": |
+          sum by (<<.GroupBy>>) (
+            container_memory_working_set_bytes{<<.LabelMatchers>>,id='/'}
+          )
+        "resources":
+          "overrides":
+            "instance":
```
olivierlemasle, line 39: It should be `node`, not `instance`.
```suggestion
            "node":
```

deploy/manifests/service-monitor.yaml
olivierlemasle: Or could we delete this ServiceMonitor, to remove the dependency to prometheus-operator?

Technically, prometheus-operator is not required to use prometheus-adapter.

olivierlemasle (commented):
Thanks @JoaoBraveCoding!
There's a small error in the `adapter-config` ConfigMap: <s>instance</s> -> node

deploy/manifests/service-monitor.yaml
olivierlemasle: Or could we delete this ServiceMonitor, to remove the dependency to prometheus-operator?

Technically, prometheus-operator is not required to use prometheus-adapter.

dgrisonnet: I meant shouldn't, the idea is to remove it to avoid having an unnecessary dependency yes.

deploy/manifests/config-map.yaml
dgrisonnet: good catch

JoaoBraveCoding: Fixed

olivierlemasle (approved):
/ok-to-test
/lgtm

dgrisonnet:
Great work, thanks @JoaoBraveCoding!

/lgtm
/approve

PR #539, opened 2022-11-09
Add initial e2e tests

olivierlemasle:
This adds E2E tests to prometheus-adapter.
These tests:
- bring up a [kind](https://kind.sigs.k8s.io/) cluster (or re-use an existing k8s cluster)
- build the prometheus-adapter image
- deploy prometheus-operator and a prometheus instance
- deploy prometheus-adapter using the manifests in [`deploy/manifests`](https://github.com/kubernetes-sigs/prometheus-adapter/tree/master/deploy/manifests)

It checks that everything can be deployed, and, as a first test, checks that:
- nodes metrics bring positive values
- there are some pod metrics

It also prints the prometheus-adapter logs.


test/e2e/e2e_test.go
```diff
@@ -0,0 +1,204 @@
...
+	"k8s.io/client-go/dynamic"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	metricsv1beta1 "k8s.io/metrics/pkg/apis/metrics/v1beta1"
+	metrics "k8s.io/metrics/pkg/client/clientset/versioned"
+)
+
+const ns = "prometheus-adapter-e2e"
+
+func initializeClient(t *testing.T) (clientset.Interface, metrics.Interface) {
```
dgrisonnet, line 41: This function is doing a lot more than just initializing clients, what do you think about splitting it a bit? Also instead of passing `t` as an argument we could return an error with smaller functions.

I am also fine with having all that stuff in a `TestMain` function

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,204 @@
...
+		t.Fatalf("Prometheus instance 'prometheus' not ready: %v", err)
+	}
+	err = waitForDeploymentReady(ctx, t, clientSet, ns, "prometheus-adapter")
+	if err != nil {
+		t.Fatalf("Deployment prometheus-adapter not ready: %v", err)
+	}
+	return clientSet, metricsClientSet
+}
+
+func waitForPrometheusReady(ctx context.Context, t *testing.T, client dynamic.Interface, namespace string, name string) error {
```
dgrisonnet, line 79: what do you think about reusing the prometheus-operator e2e test framework for that? https://github.com/prometheus-operator/prometheus-operator/blob/1dfecab7c2706f3591d6bb6d5eb8f6b2518f1b45/test/framework/prometheus.go#L309-L320

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,204 @@
...
+		return false, nil
+	})
+}
+
+func TestNodeMetrics(t *testing.T) {
+	_, metricsClient := initializeClient(t)
+
+	ctx := context.Background()
+	var nodeMetrics *metricsv1beta1.NodeMetricsList
+	err := wait.PollImmediateWithContext(ctx, 5*time.Second, 120*time.Second, func(ctx context.Context) (bool, error) {
```
dgrisonnet, line 156: The e2e cluster should be ready by that time so we should do more active polling with a lower timeout

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,204 @@
...
+		assert.Positivef(t, positiveCpu, "CPU usage for node %s is %v, should be > 0", nodeMetric.Name, nodeMetric.Usage.Cpu())
+	}
+}
+
+func TestPodMetrics(t *testing.T) {
+	_, metricsClient := initializeClient(t)
+
+	ctx := context.Background()
+	var podMetrics *metricsv1beta1.PodMetricsList
+	err := wait.PollImmediateWithContext(ctx, 5*time.Second, 120*time.Second, func(ctx context.Context) (bool, error) {
```
dgrisonnet, line 184: Same comment wrt to the polling interval and timeout duration post setup

test/prometheus-manifests/service-monitor-kubelet.yaml
```diff
@@ -0,0 +1,101 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
```
dgrisonnet, line 2: Could you make this ServiceMonitor more minimal? We shouldn't need all the relabeling rules here and the only path that we want to scrape on Kubelet is `/metrics/resources`.

test/prometheus-manifests/prometheus.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: prometheus-adapter-e2e
+spec:
+  replicas: 2
+  serviceAccountName: prometheus
+  serviceMonitorSelector: {}
+  alerting:
+    alertmanagers:
+      - namespace: default
+        name: alertmanager
+        port: web
```
dgrisonnet, line 14: we don't need alerting

test/e2e/e2e_test.go
olivierlemasle: I ended using a `TestMain` function.

test/e2e/e2e_test.go
olivierlemasle: That's a good idea. I tried but got stuck in dependency hells, as prometheus-operator brings Kubernetes, prometheus and opentelemetry dependencies in different versions than prometheus-adapter.
However, I could use [prometheus-operator client](https://pkg.go.dev/github.com/prometheus-operator/prometheus-operator/pkg/client) (which is a different module with fewer dependencies) to make this function less verbose, if you think it's better.

test/prometheus-manifests/service-monitor-kubelet.yaml
olivierlemasle: I made the ServiceMonitor more minimal. However, I'm still scraping `/metrics/cadvisor`, not `/metrics/resource`, which do not provide `container_cpu_usage_seconds_total{id='/'}`.

test/prometheus-manifests/prometheus.yaml
olivierlemasle: Oops, removed.

test/prometheus-manifests/service-monitor-kubelet.yaml
dgrisonnet: It was renamed to `container_cpu_usage_seconds` in https://github.com/kubernetes/kubernetes/pull/86282. Since `/metrics/resource` is a more lightweight version of `/metrics/cadvisor`, it would be better to use the new endpoint and the new metric.

test/e2e/e2e_test.go
dgrisonnet: > That's a good idea. I tried but got stuck in dependency hells, as prometheus-operator brings Kubernetes, prometheus and opentelemetry dependencies in different versions than prometheus-adapter.

Oh that's unfortunate, I am fine with the current approach, just though that it could maybe be simplified by importing existing code.

> However, I could use [prometheus-operator client](https://pkg.go.dev/github.com/prometheus-operator/prometheus-operator/pkg/client) (which is a different module with fewer dependencies) to make this function less verbose, if you think it's better.

Yeah that's a good idea

olivierlemasle: I refactored the tests to use the prometheus-operator client.

test/prometheus-manifests/service-monitor-kubelet.yaml
olivierlemasle: Actually, https://github.com/kubernetes/kubernetes/pull/86282 was kind of reverted by https://github.com/kubernetes/kubernetes/pull/89540 in Kubernetes 1.19, so the metric is named `container_cpu_usage_seconds_total` (for k8s < 1.18 and k8s >= 1.19), and that's the metric [documented](https://kubernetes.io/docs/reference/instrumentation/metrics/).

What I meant is that, using the path `/metrics/resource`, query`container_cpu_usage_seconds_total{id='/'}` returns nothing, because the label `id` is never empty (however it works on path `/metrics/cadvisor`).

So I've updated the configmap manifest to use `node_cpu_usage_seconds_total` and `node_memory_working_set_bytes` instead. I can make a separate PR for that if you prefer. I was unsure because you suggested yourself in https://github.com/kubernetes-sigs/prometheus-adapter/pull/531#discussion_r961844262 to use `{id='/'}` for node queries.

deploy/manifests/config-map.yaml
```diff
@@ -13,7 +13,7 @@ data:
         "nodeQuery": |
           sum by (<<.GroupBy>>) (
             irate(
-                container_cpu_usage_seconds_total{<<.LabelMatchers>>,id='/'}[4m]
+                node_cpu_usage_seconds_total{<<.LabelMatchers>>}[4m]
```
dgrisonnet, line 16: To do that we would need to install node-exporter. Although this is what is done in most setup these days, I would rather limit the amount of dependencies of this repo to just Prometheus and Kubernetes

test/prometheus-manifests/service-monitor-kubelet.yaml
dgrisonnet: > What I meant is that, using the path /metrics/resource, querycontainer_cpu_usage_seconds_total{id='/'} returns nothing, because the label id is never empty (however it works on path /metrics/cadvisor

Oh that's interesting, I wasn't aware that `id` wasn't included in the metric exposed `/metrics/resource`. I guess an alternative could then be to group by `node` label since it is injected by prometheus-operator, but at this point I am not really sure what's best between that and depending on node-exporter. For sure we should move away from /metrics/cadvisor since it is bound to disappear.

dgrisonnet:
Thinking about it again, let's move away from /metrics/cadvisor for node metrics and rely on node-exporter instead. That's what most of the community is doing and we've already had asks to switch.

So for the tests we would want:
1. ServiceMonitor on kubelet /metrics/resource for CPU queries
2. ServiceMonitor on node-exporter for Node queries

olivierlemasle:
@dgrisonnet I'm not sure I understand the issue with the way it works in this PR.

`node_cpu_usage_seconds_total` is provided by the kubelet (cf https://github.com/kubernetes/kubernetes/blob/64af1adaceba4db8d0efdb91453bce7073973771/pkg/kubelet/metrics/collectors/resource_metrics.go#L30), so we dont't need node-exporter, do we?

That being said, I guess that prometheus-adapter is often used in conjunction with node-exporter (in kube-prometheus or openshift's cluster-monitoring-operator), so I'm not against using node-exporter in default manifests and in E2E.

dgrisonnet:
:weary: I totally forgot that node-level metrics were introduced sorry about that.

IIRC node-exporter and kubelet are getting the data from the same place so reducing the dependencies to just kubelet would definitely be better.

dgrisonnet (approved):
Thank you for bearing with me on this one. As a follow-up we should enable the tests in CI.

/lgtm
/approve

olivierlemasle:
Thank you @dgrisonnet :tada: 

I've just updated https://github.com/kubernetes/test-infra/pull/27948

PR #540, opened 2022-11-09
Use golangci-lint

olivierlemasle:
Since Go 1.19, `gofmt` [formats comments](https://go.dev/doc/go1.19#go-doc). This caused `make verify` fail when using Go 1.19 with some license header comments.

`gofmt` output format also changes with Go 1.19, so `hack/gofmt-all.sh` is modified to be more robust. I took this opportunity to slighty improve this script.

Also, I modified `Makefile`:
- `verify` should not depend of `test` (it should be static analysis only)
- `verify-generated` should depend of `update-generated` (otherwise, it always succeeds)

---

PR updated: use golangci-lint


dgrisonnet:
Same comment as in https://github.com/kubernetes-sigs/custom-metrics-apiserver/pull/111#issuecomment-1323555943, I'd very much prefer if we used golangci-lint.

olivierlemasle:
PR updated to use golangci-lint.

In the `Makefile`, two changes are made in addition to adding golangci-lint:
- target `verify` does not depend on `test` any more;
- target `verify-generated` depends on `update-generated`.

Most other changes are formatting / style / naming.

Notable changes are:
- Adding `MinVersion: tls.VersionTLS12` to `TLSClientConfig`.
  Current code was reported by [gosec](https://github.com/securego/gosec) as **G402** (CWE-295): _TLS MinVersion too low_
  I used `MinVersion: tls.VersionTLS12` because it's what client-go [uses](https://github.com/kubernetes/client-go/blob/1ac8d459351e21458fd1041f41e43403eadcbdba/transport/transport.go#L92).
- In `pkg/client/api.go`, [`req.WithContext(ctx)`](https://github.com/kubernetes-sigs/prometheus-adapter/blob/fdfecc8d7f66cde4dc2c85858773463911819730/pkg/client/api.go#L68) did not set the request's context, because `WithContext` is a pure function.  This is fixed by using `http.NewRequestWithContext`.

dgrisonnet:
Thank you @olivierlemasle, the PR looks good to me overall. Just a small nitpick, would you mind doing the TLS `MinVersion` change in a separate PR? It is important enough in my opinion to have its own changelog entry. In the meantime, you can silence the report in this PR.

olivierlemasle:
@dgrisonnet Done.
I'll create the follow-up PR with the TLS change after this is merged.

dgrisonnet:
Thanks!

/lgtm
/approve

PR #546, opened 2022-12-05
Refactor adding logging flags

olivierlemasle:
Instead of using `logs.AddGoFlags(local)` to add logging flags to Go `flags.FlagSet`, then adding them to `github.com/spf13/pflag.FlagSet`, we directly add them to `github.com/spf13/pflag.FlagSet`.

/kind cleanup


dgrisonnet (commented):
/lgtm
/approve

Issue #545, opened 2022-12-04
Custom metrics: use the timestamp from sample


olivierlemasle:
For custom metrics, the timestamp is `time.Now()`; a `//TODO` was added 4 years ago to use the right timestamp instead:

https://github.com/kubernetes-sigs/prometheus-adapter/blob/89584579687e4a66f352a1ba26e957951de628e7/pkg/custom-provider/provider.go#L101-L102

We should use the timestamp from the sample, as it is done for external metrics and resource metrics.

impact-maker:
@olivierlemasle Hi, I have been working on this issue and submited a pull request #554 for the same. I am a newly introduced developer for this organisation and would need some inputs to contribute for the same. Kindly let me know if there is any guide to read before the contributions or any rules to look for testing the code for the environment.

olivierlemasle:
Thank you and welcome @impact-maker 

First, you'll find the Kubernetes Contributing Guidelines in [CONTRIBUTING.md](https://github.com/kubernetes-sigs/prometheus-adapter/blob/master/CONTRIBUTING.md). It also includes a guide on how to sign the CLA, which is required for any contribution, and general guidelines (how to write a good commit message, etc.). https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md is also a useful general guide.

From a technical perspective, for prometheus-adapter, you can use `make verify test` to run the linters / checkers and execute the tests locally, before pushing the code to GitHub.
You can use `make container` to build a container image of prometheus-adapter and test if everything is ok.

I'll review your PR later today.

impact-maker:
@olivierlemasle  Thank you for guiding me through this process. I have tried to do a fair bit of research to go through the documentations provided. 
I have made another commit in the last pull request. Kindly review that and help me cut through the steep learning curve of contributing to this project. 
I have tried my best to run make verify test command but failed to do so and I am still not able to compile/ test the code before commiting. Your inputs on the same would really help me provide useful code contributions for this issue.

PR #554, opened 2023-01-03
Changed the timestamp without testing

impact-maker:
Made changes to timestamp. Currently it is not tested and will require further testing

### DIFF
--- a/pkg/custom-provider/provider.go
+++ b/pkg/custom-provider/provider.go
@@ -98,8 +98,20 @@ func (p *prometheusProvider) metricFor(value pmodel.SampleValue, name types.Name
 		Metric: custom_metrics.MetricIdentifier{
 			Name: info.Metric,
 		},
+		
 		// TODO(directxman12): use the right timestamp
-		Timestamp: metav1.Time{Time: time.Now()},
+		
+		
+		/*Making the changes in the timestamp as per issue
+		number #545*/
+
+		// Previously set timestamp
+		//Timestamp: metav1.Time{Time: time.Now()},
+
+		// Newly set timestamp
+		Timestamp: metav1.Time{
+			Time: value.Timestamp.Time(),
+		},
 		Value:     *q,
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/custom-provider/provider.go
```diff
@@ -98,8 +98,20 @@ func (p *prometheusProvider) metricFor(value pmodel.SampleValue, name types.Name
...
-		Timestamp: metav1.Time{Time: time.Now()},
+		
+		
+		/*Making the changes in the timestamp as per issue
+		number #545*/
+
+		// Previously set timestamp
+		//Timestamp: metav1.Time{Time: time.Now()},
+
+		// Newly set timestamp
```
olivierlemasle, line 111: Please remove these comments. This kind of explanation will be in the commit message, not in the code.

pkg/custom-provider/provider.go
```diff
@@ -98,8 +98,20 @@ func (p *prometheusProvider) metricFor(value pmodel.SampleValue, name types.Name
...
+		Timestamp: metav1.Time{
+			Time: value.Timestamp.Time(),
```
olivierlemasle, line 113: This will not compile.
`value` is a [`github.com/prometheus/common/model.SampleValue`](https://pkg.go.dev/github.com/prometheus/common/model#SampleValue), which has no `Timestamp` field.
You need to change `metricFor` to take a [`github.com/prometheus/common/model.Sample`](https://pkg.go.dev/github.com/prometheus/common/model#Sample) or a struct which includes both `Value` and `Timestamp`.
To do that, you'll need to refactor every function which calls `metricFor`.

olivierlemasle (changes requested):
Thanks for the PR @impact-maker.

In addition to the comments here and https://github.com/kubernetes-sigs/prometheus-adapter/issues/545#issuecomment-1369739034, you'll need to sign the CNCF CLA before contributing your patch.

impact-maker (commented):
Added the sample model class to retrieve the timestamp.

pkg/custom-provider/provider.go
impact-maker: @olivierlemasle I have refactored the metricFor function in the provider.go class to add the Sample model class.

pkg/custom-provider/provider.go
impact-maker: Done in the latest commit .

PR #559, opened 2023-01-25
Fix yaml in sample config & docs

asherf:
new here. so this may be wrong. but as far as I can tell, those commas make the yaml invalid....
at least this is what I saw when I ran Prometheus Adapter with that config.


olivierlemasle (approved):
Thank you @asherf

/ok-to-test

dgrisonnet (approved):
/approve

PR #564, opened 2023-02-16
selector -> labelSelector

stoyansbg:
Reference: https://github.com/kubernetes/design-proposals-archive/blob/main/instrumentation/custom-metrics-api.md#relationship-to-hpa-v2


stoyansbg:
CLA signed

PR #589, opened 2023-06-22
Add support for OpenAPI v3

dgrisonnet:


cmd/adapter/adapter.go
```diff
@@ -292,15 +294,23 @@ func main() {
 	}
 	cmd.Name = "prometheus-metrics-adapter"
 
-	cmd.OpenAPIConfig = genericapiserver.DefaultOpenAPIConfig(generatedopenapi.GetOpenAPIDefinitions, openapinamer.NewDefinitionNamer(api.Scheme, customexternalmetrics.Scheme))
```
dgrisonnet, line 295: flag initialization need to happen before the OpenAPI configuration since it can be set by the users via a CLI flag.

PR #599, opened 2023-08-17
Fix metric unregistered

bogo-y:


pkg/client/metrics/metrics.go
```diff
@@ -23,25 +23,27 @@ import (
...
-			Buckets: prometheus.ExponentialBuckets(0.0001, 2, 10),
+			Buckets: prometheus.ExponentialBuckets(0.001, 2, 10),
 		},
 		[]string{"endpoint", "server"},
 	)
 )
 
 func init() {
-	prometheus.MustRegister(queryLatency)
+	legacyregistry.MustRegister(queryLatency)
```
dgrisonnet, line 46: We shouldn't add metrics to the Kubernetes global registry. We should rather dependency inject the Prometheus registry like it was done in the past in metrics-server: https://github.com/kubernetes-sigs/metrics-server/pull/587/. There was a few changes made to the original work in https://github.com/kubernetes-sigs/metrics-server/pull/634/, but I think we should follow the same process in prometheus-adapter.

pkg/client/metrics/metrics.go
```diff
@@ -23,25 +23,27 @@ import (
...
 
 var (
 	// queryLatency is the total latency of any query going through the
 	// various endpoints (query, range-query, series).  It includes some deserialization
 	// overhead and HTTP overhead.
-	queryLatency = prometheus.NewHistogramVec(
-		prometheus.HistogramOpts{
+	queryLatency = metrics.NewHistogramVec(
+		&metrics.HistogramOpts{
 			Name:    "cmgateway_prometheus_query_latency_seconds",
```
dgrisonnet, line 37: Since this metric was never expose, we can safely change its name without breaking anyone. I would suggest the following one:

```suggestion
			Namespace:    "prometheus_adapter",
			Subsystem:    "prometheus_client",
			Name:    "request_duration_seconds",
```

pkg/client/metrics/metrics.go
```diff
@@ -23,25 +23,27 @@ import (
...
 			Help:    "Prometheus client query latency in seconds.  Broken down by target prometheus endpoint and target server",
-			Buckets: prometheus.ExponentialBuckets(0.0001, 2, 10),
+			Buckets: prometheus.ExponentialBuckets(0.001, 2, 10),
```
dgrisonnet, line 39: I think we could use the default buckets: https://github.com/prometheus/client_golang/blob/51d24f86807c24ce866355d0c82258420f571c17/prometheus/histogram.go#L264

dgrisonnet (commented):
Great find @bogo-y!

bogo-y:
Thanks for your opinion, and I'm looking forward to your next review. @dgrisonnet

pkg/client/metrics/metrics.go
```diff
@@ -18,30 +18,45 @@ package metrics
...
-			Buckets: prometheus.ExponentialBuckets(0.0001, 2, 10),
+	queryLatency = metrics.NewHistogramVec(
+		&metrics.HistogramOpts{
+			Namespace: "prometheus_adapter",
+			Subsystem: "prometheus_client",
+			Name:      "request_duration_seconds",
+			Help:      "Prometheus client query latency in seconds.  Broken down by target prometheus endpoint and target server",
+			Buckets:   prometheus.DefBuckets,
 		},
 		[]string{"endpoint", "server"},
```
simonpasquier, line 45: `endpoint` is already a target label. If `honor_labels` is false (which is the default), the metric label will be renamed to `exported_endpoint` to preserve the target label.
Since the metric was never exposed before, I'd suggest using `path` instead.
```suggestion
		[]string{"path", "server"},
```

dgrisonnet: yeah path would be a good fit since the values can be one of these: https://github.com/kubernetes-sigs/prometheus-adapter/blob/master/pkg/client/api.go#L139-L141

simonpasquier (commented):
/lgtm

cmd/adapter/adapter.go
```diff
@@ -273,6 +273,12 @@ func (cmd *PrometheusAdapter) addResourceMetricsAPI(promClient prom.Client, stop
 		return err
 	}
 
+	metricsHandler, err := mprom.MetricsHandler()
+	if err != nil {
+		return err
+	}
+	server.GenericAPIServer.Handler.NonGoRestfulMux.HandleFunc("/metrics", metricsHandler)
```
dgrisonnet, line 280: I haven't tested it yet myself, but I would expect this to cause a [panic](https://cs.opensource.google/go/go/+/refs/tags/go1.21.2:src/net/http/server.go;l=2517-2519) since `/metrics` has already a handler by default unless we set `EnabledMetrics = false` in the server config like in https://github.com/kubernetes-sigs/metrics-server/blob/f792d7022c270df795adda1de8bde7b1d4970c20/pkg/server/config.go#L74.

bogo-y:
@dgrisonnet I can't find out what caused the e2e test failed. I don't know why the changes can influence the inherent features. Could you please give me more guidance?

bogo-y:
@dgrisonnet I figured out the problem caused e-2-e test faild, so please take a look again, thx.

bogo-y:
@dgrisonnet /ping

dgrisonnet (approved):
/lgtm
/approve

dgrisonnet:
Thank you for fixing this @bogo-y and sorry for the delay.

PR #618, opened 2023-10-30
Add a toggle to disable HTTP/2 on the server to mitigate CVE-2023-44487

machine424:
until the Go standard library and golang.org/x/net are fully fixed.

---

Manually tested using `curl -sI --http2 -k -w "%{http_version}\n" -o /dev/null https://localhost:6443` (after changing the base image to get `curl`). It returns `1.1` when `--disable-http2` is set.

I could add an e2e test but it'll take some time.

### DIFF
--- a/cmd/adapter/adapter.go
+++ b/cmd/adapter/adapter.go
@@ -82,7 +82,8 @@ type PrometheusAdapter struct {
 	MetricsRelistInterval time.Duration
 	// MetricsMaxAge is the period to query available metrics for
 	MetricsMaxAge time.Duration
-
+	// DisableHTTP2 indicates that http2 should not be enabled.
+	DisableHTTP2  bool
 	metricsConfig *adaptercfg.MetricsDiscoveryConfig
 }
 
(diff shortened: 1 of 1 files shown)


cmd/adapter/adapter.go
```diff
@@ -357,6 +360,14 @@ func main() {
 		klog.Fatalf("unable to install resource metrics API: %v", err)
 	}
 
+	// disable HTTP/2 to mitigate CVE-2023-44487 until the Go standard library
+	// and golang.org/x/net are fully fixed.
+	server, err := cmd.Server()
+	if err != nil {
+		klog.Fatalf("unable to fetch server: %v", err)
+	}
+	server.GenericAPIServer.SecureServingInfo.DisableHTTP2 = cmd.DisableHTTP2
```
machine424, line 369: do you want to add an info log for it?

dgrisonnet: I don't think it is necessary since the information is already in the adapter config

dgrisonnet (approved):
/lgtm
/approve

PR #645, opened 2024-02-29
Instrument metrics 

cerberus20:
image


cerberus20:
@dgrisonnet  /ping

cerberus20:
@dgrisonnet Can you please help review? thank you.

pkg/client/metrics/metrics.go
```diff
@@ -44,15 +44,35 @@ var (
...
+
+	// define a counter for API errors for various ErrorTypes
+	apiErrorCount = metrics.NewCounterVec(
+    		&metrics.CounterOpts{
+        		Namespace: "prometheus_adapter",
+        		Subsystem: "prometheus_client",
+        		Name:      "api_errors_total",
+        		Help:      "Total number of API errors",
+    		},
+    		[]string{"error_code", "path", "server"},
```
dgrisonnet, line 58: In order to compute an error rate and produce an alert out of this metric, it is usually better to record all the request including the one that succeed.

A common practice is to have a `http_requests_total` metric with a `code` label. In you case I would suggest to have a `api_requests_total` metric that record all the requests and rename the label to `code` and set it to the http status code of the request.

pkg/client/metrics/metrics.go
```diff
@@ -44,15 +44,35 @@ var (
...
+
+	errRegisterQueryLatency := registry.Register(queryLatency)
+	if errRegisterQueryLatency != nil {
+		return nil, errRegisterQueryLatency
+	}
+
+	errRegisterAPIErrorCount := registry.Register(apiErrorCount)
+	if errRegisterAPIErrorCount != nil {
+		return nil, errRegisterAPIErrorCount
 	}
```
dgrisonnet, line 74: Since we don't necessarily need to differentiate between the errors here, it is usually a common practice in Go to just have one `err` variable and reuse it for reach `Register` call.

pkg/client/metrics/metrics.go
cerberus20: ack, makes sense, updated with api_requests_total. thanks for the suggestion.

pkg/client/metrics/metrics.go
cerberus20: updated, thanks

pkg/client/metrics/metrics.go
```diff
@@ -69,20 +87,24 @@ type instrumentedGenericClient struct {
...
-		// skip calls where we don't make the actual request
 		if err != nil {
-			if _, wasAPIErr := err.(*client.Error); !wasAPIErr {
-				// TODO: measure API errors by code?
-				return
+			if apiErr, wasAPIErr := err.(*client.Error); wasAPIErr {
+				apiRequestsTotal.With(prometheus.Labels{"code": string(apiErr.Type), "path": endpoint, "server": c.serverName}).Inc()
+			} else {
+				apiRequestsTotal.With(prometheus.Labels{"code": "generic", "path": endpoint, "server": c.serverName}).Inc()
 			}
```
dgrisonnet, line 100: When I first looked at your PR I had forgotten that the client errors were wrapped https://github.com/kubernetes-sigs/prometheus-adapter/blob/master/pkg/client/types.go#L27-L31 which is not ideal as it is less explicit to users than a normal http code would be. I feel like we should change some things in order to surface the http code and expose it in the metrics.

I'll be out till the 27th of May, but I'll try to have a look at that again when I am back.

cerberus20: @dgrisonnet  I updated code to add HTTPStatusCode to APIResponse to expose the http code in the metric. Can you please review? Thank you.

Issue #638, opened 2024-02-11
Improve security hygiene and documentation


ricardoapl:
I noticed prometheus-adapter is lacking in security documentation. I'm thinking about something similar to what was done in https://github.com/kubernetes/kube-state-metrics/issues/2274 for Security Slam - Kubernetes Lightning Round.

Would you like to do this for prometheus-adapter? This means:
- Generating an SBOM automatically when a new release is tagged, so that users have a better understanding of whether or not they're affected by a particular vulnerability
- Generating SLSA/provenance attestation automatically for each new release, so that users can verify the authenticity and trustworthiness of the build process
- Initializing VEX feed and generating OpenVEX data on each release, so that users have a better understanding of which vulnerabilities present a genuine risk
- Setting up CLOMonitor tracking and following-up on any security checks

I believe this relates to https://github.com/kubernetes/sig-release/blob/193a3cdf8d73e0888c7f6829eea3716918a5af4a/roadmap.md

ricardoapl:
Maybe we can close this since prometheus-adapter will be archived?

/cc @dgrisonnet

dgrisonnet:
Yes :+1:

PR #646, opened 2024-03-02
WIP - Generate sbom on release

ricardoapl:
Generate an SBOM for each new release of prometheus-adapter

Part of https://github.com/kubernetes-sigs/prometheus-adapter/issues/638

### DIFF
--- a/.github/workflows/sbom.yaml
+++ b/.github/workflows/sbom.yaml
@@ -1,0 +1,41 @@
+name: Generate SBOM with Kubernetes BOM
+
+on:
+  release:
+    types:
+      - published
+
+permissions:
+  contents: read
+
+jobs:
+  sbom:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    env:
+      OUTPUT: prometheus-adapter-${{ github.ref_name }}.spdx
+      TAG: ${{ github.ref_name }}
+
+    steps:
+      - name: Fetch source code into GITHUB_WORKSPACE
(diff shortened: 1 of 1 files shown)


.github/workflows/sbom.yaml
```diff
@@ -0,0 +1,41 @@
+name: Generate SBOM with Kubernetes BOM
+
+on:
+  release:
+    types:
+      - published
+
+permissions:
+  contents: read
```
ricardoapl, line 9: GitHub recommends setting minimum token permissions for the `GITHUB_TOKEN`. Although unlikely, there is a chance that when a new job is added to this workflow, its job permissions could be left undefined because of human error. We can prevent that by setting permissions definitions as read-only at the [top level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions).

ricardoapl:
It's not clear to me how we should go about signing SBOMs. [KEP-3031: Signing release artifacts](https://github.com/kubernetes/enhancements/tree/master/keps/sig-release/3031-signing-release-artifacts) doesn't seem to apply outside of k/k, and there's also an open discussion at https://github.com/kubernetes/release/issues/2286. I suggest we tackle signing SBOMs in a separate pull request.

What do you think?

ricardoapl:
Maybe not worth pursuing this given that prometheus-adapter will be archived?

ricardoapl:
/close

This is not relevant anymore as discussed in https://github.com/kubernetes-sigs/prometheus-adapter/issues/638

PR #685, opened 2024-12-02
Upgrade Go version to 1.22.9 to fix CVEs

heshamelsherif97:
Upgrades Go version to latest patch version 1.22.9
Which includes multiple CVEs fixes


slashpai:
https://github.com/kubernetes-sigs/prometheus-adapter/pull/700 is opened to move to 1.24.3, it needs test config updates, awaiting review

PR #698, opened 2025-05-06
fix(docs): correct malformed YAML keys in sample-config.yaml

### LINKED ISSUE #697: Quoting issues in example config YAML
The `docs/sample-config.yaml` is not usable as it is currently. Issues arises due to doublequote-quoted strings in the YAML that contain other double-quotes, leading to parsing errors.

When attempting to start the adapter with the command:
```
./adapter --config=docs/sample-config.yaml
```
The following error is encountered:
```
I0506 11:35:38.062439 3062873 adapter.go:113] successfully using in-cluster auth
F0506 11:35:38.062573 3062873 adapter.go:340] unable to load metrics discovery config: unable to load metrics discovery configuration: unable to parse metrics discovery config: yaml: line 54: did not find expected key
```
Additional details from `yamlfix`:
```
ruyaml.parser.ParserError: while parsing a block mapping
  in "<unicode string>", line 13, column 3:
    - seriesQuery: '{__name__=~"^conta ...
      ^ (line: 13)
expected <block end>, but found '<scalar>'
  in "<unicode string>", line 30, column 70:
     ... {<<.LabelMatchers>>,container!="POD"}[2m])) by (<<.GroupBy>>)"
                                         ^ (line: 30)
```

Seems like there are multiple occasions of those quoting issue in the values for `metricsQuery`, for example here:
https://github.com/kubernetes-sigs/prometheus-adapter/blob/01919d0ef11859bc214e0c8a8bd5368afd9d47f7/docs/sample-config.yaml#L30

skoenig:
Fixes: #697 
Replace double quotes with single quotes in metricsQuery to avoid parsing errors.

### DIFF
--- a/docs/sample-config.yaml
+++ b/docs/sample-config.yaml
@@ -27,7 +27,7 @@ rules:
   # This is a Go template where the `.Series` and `.LabelMatchers` string values
   # are available, and the delimiters are `<<` and `>>` to avoid conflicts with
   # the prometheus query language
-  metricsQuery: "sum(rate(<<.Series>>{<<.LabelMatchers>>,container!="POD"}[2m])) by (<<.GroupBy>>)"
+  metricsQuery: 'sum(rate(<<.Series>>{<<.LabelMatchers>>,container!="POD"}[2m])) by (<<.GroupBy>>)'
 
 # this rule matches cumulative cAdvisor metrics not measured in seconds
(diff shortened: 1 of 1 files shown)


richabanker:
@dgrisonnet seems like a simple change. Waiting on your approval

TheMeier (approved):

PR #703, opened 2025-08-11
chore: Deprecate Prometheus Adapter

### LINKED ISSUE #701: Deprecate Prometheus Adapter
We want to deprecate the project in the near-future, post-triage, and suggest users to fulfill their use-cases using KEDA.

Additionally, [unlike `metrics-server`](https://keda.sh/docs/2.17/scalers/cpu/#:~:text=cpu%20and%20memory%20metrics%20from%20the%20Kubernetes%20Metrics%20Server), KEDA does not need Prometheus Adapter as a prerequisite, and is able to fairly facilitate the Discovery-Associate-Naming-Querying loop as a replacement.

I'll start work on a migration document soon, in addition to making sure the proposed replacement delivers on all supported use-cases here.

- [x] Ensure all PA use-cases are (more-or-less) accomplished using KEDA.
- [x] Prepare migration document.
- [ ] Get the rest of the SIG to review the document so we're on the same page.
- [ ] Engage with the community and get their opinions on this.
- [ ] Based on the feedback, announce the deprecation on the mailing list, or, help folks that are interested in maintaining the repository if we (the community) decide overwhelmingly against it.

Do note that the SIG lacks incentives to maintain the already superseded repository, so folks from the community will need to step up if its the latter case.

Also note that we've already consulted with appropriate upstream and community organizations about the possibility of adoption, but the general consensus we received was to deprecate this.

cc @kubernetes-sigs/instrumentation-admins

***

### TODOs

- [ ] Prepend deprecation notice in the README.md

rexagod:
Push KEDA's Prometheus, CPU, and Memory scalers instead.

Fixes: #701

### DIFF
--- a/DEPRECATION.md
+++ b/DEPRECATION.md
@@ -1,0 +1,292 @@
+## Deprecation Notice for Prometheus Adapter
+
+#### Users are advised to follow this document to migrate from (the soon-to-be deprecated) Prometheus Adapter to KEDA.
+
+### Deprecation Rationale
+
+Please refer to the original [deprecation tracker](https://github.com/kubernetes-sigs/prometheus-adapter/issues/701) for the discussion around the deprecation of Prometheus Adapter.
+
+### Deprecation Notice
+
+Everything encompassed by this repository is being deprecated. This includes the Prometheus Adapter itself, all its components, and any related configurations or scripts that are part of this repository. Note that any assets related to Prometheus Adapter that are not part of this repository are not covered by this deprecation notice.
+
+### Deprecation Timeline
+
+The SIG plans to deprecate Prometheus Adapter in the next release cycle, which is expected to be around a couple months from now. After this period, Prometheus Adapter will no longer receive updates or support.
+
+### Deprecation Alternative: KEDA
+
+Below are correlations between the various configuration fields offered by Prometheus Adapter, and their equivalent interpretations using the configuration options offered by KEDA.
+
+However, please keep in mind that:
+
+- The examples below assume you are scaling a single Deployment named `my-app` in namespace `default`, and that your Prometheus instance is reachable at `http://prometheus.monitoring.svc:9090`.
+- KEDA's query blocks do not support the templating syntax used in Prometheus Adapter. Instead, you need to write concrete queries that return a single number (instant vector with a single element or a scalar).
+  - The queries should be adapted to your environment, and you can use label selectors to narrow down the metrics to the specific pods or namespaces you are interested in (for example, `pod=~"my-app-.*"` and a fixed namespace).
+- Unlike Prometheus Adapter, KEDA does not support defining [custom metrics](https://kubernetes.io/docs/reference/external-api/custom-metrics.v1beta2/), and as such, all newly mapped metrics for HPA consumption on KEDA's end are exposed as [external metrics](https://kubernetes.io/docs/reference/external-api/external-metrics.v1beta1/).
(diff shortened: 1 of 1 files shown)


DEPRECATION.md
```diff
@@ -0,0 +1,292 @@
...
+
+Below are correlations between the various configuration fields offered by Prometheus Adapter, and their equivalent interpretations using the configuration options offered by KEDA.
+
+However, please keep in mind that:
+
+- The examples below assume you are scaling a single Deployment named `my-app` in namespace `default`, and that your Prometheus instance is reachable at `http://prometheus.monitoring.svc:9090`.
+- KEDA's query blocks do not support the templating syntax used in Prometheus Adapter. Instead, you need to write concrete queries that return a single number (instant vector with a single element or a scalar).
+  - The queries should be adapted to your environment, and you can use label selectors to narrow down the metrics to the specific pods or namespaces you are interested in (for example, `pod=~"my-app-.*"` and a fixed namespace).
+- Unlike Prometheus Adapter, KEDA does not support defining [custom metrics](https://kubernetes.io/docs/reference/external-api/custom-metrics.v1beta2/), and as such, all newly mapped metrics for HPA consumption on KEDA's end are exposed as [external metrics](https://kubernetes.io/docs/reference/external-api/external-metrics.v1beta1/).
```
a7i, line 26: IMO, this seems like a blocker for deprecating it as we are using it for custom metrics

TheMeier: `custom-metrics.v1beta2` is BETA! what prevents you from to switching to `external-metrics.v1beta1`?

a7i: @TheMeier In my use-case, I'm using Datadog as the external metrics provider

TheMeier: @a7i did you raise an issue with them?

dgrisonnet:
I will put that on top of my backlog

/remove-lifecycle rotten

DEPRECATION.md
rexagod: External metrics represent external endpoints and non-objects, whereas custom metrics represent objects in the cluster. It is possible, however, to use external metrics to query for objects too, making it a possible superset, in terms of potential (but not recommended) use cases. This is not something we recommend as it demands putting in brittle selectors for objects into queries, so that external metrics could end up doing what custom metrics did as well, which, IIUC, also seems to be the case KEDA implicitly promotes by only supporting external metrics.

You can either try to do the same, as a short-term fix, in Datadog, or ask them to support custom metrics too.

rexagod: Folks reading this, please do let us know here if this is blocking for you, so that we may better scope the reliance of downstream solutions (or lack thereof) on Prometheus Adapter, and reassess the future steps here.

moreiravictor: What blocks us here is not being able to set custom metric names (keda set s-{index}-prometheus) cause we need to map the scaling events based on custom names

kevholmes: We are migrating a fleet from DataDog while keeping its autoscaling path authoritative. The DataDog Cluster Agent already owns the cluster scoped `v1beta1.external.metrics.k8s.io` APIService.

We're moving users to Prometheus Adapter's `custom.metrics.k8s.io` API which allows both providers to coexist at the same time as teams migrate on their own timelines during the migration period.

Is there any way a custom-metrics capable successor could coexist with an existing external-metrics provider?

